Issue #18, opened 2014-12-08
Failure when using cerner_splunk_forwarder_monitors inside a separate run_context


tboettch:
When the ```cerner_splunk_forwarder_monitors``` resource is used inside a run_context other than the top-level context (e.g. from a provider that declares ```use_inline_resources```) it is unable to send notifications to the "splunk" service because that Chef resource is not visible. To avoid this, the ```cerner_splunk_forwarder_monitors``` should declare a copy of the service resource.

acharlieh:
More design thoughts: https://github.com/cerner/cerner_splunk/pull/22#issuecomment-124629692

PR #22, opened 2014-12-11
Issue 18: Declare Splunk service inside resource provider.

tboettch:
Replaces #19 to use stable branch. This addresses #18.

### DIFF
--- a/definitions/splunk_service.rb
+++ b/definitions/splunk_service.rb
@@ -1,0 +1,15 @@
+# Declares the Splunk service resource, performing the specified action.
+define splunk_service, resource_name: nil, action: :nothing do
+  if platform_family?('windows')
+    service_name = 'SplunkForwarder'
+  else
+    service_name = 'splunk'
+  end
+
+  resource_name = params[:resource_name] || params[:name]
+  service resource_name do
+    service_name service_name
+    action params[:action]
+    supports status: true, start: true, stop: true, restart: true
+  end
+end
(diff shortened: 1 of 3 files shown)


mattnelson:
I also ran into this issue when I took foodcritic's suggestion to declare `use_inline_resources` in my LWRP. Is the rebase the only issue holding this up?

http://www.foodcritic.io/#FC017

acharlieh:
So the rebase was holding it up... but as I got to thinking about it again after your comment @mattnelson, I think this also isn't the correct solution.

Consider: https://github.com/acharlieh/test-lwrp Which I believe is a simplified version of the state of things with the proposed solution.

If we just add a copy of the service resource (in this case a `bash[foo]` resource) to the forwarder_monitors LWRP (the default LWRP), when we're in a `use_inline_resources` (the inline LWRP), the restart (run) action is executed at the end of the embedded chef run... Which means as we add calls to similar LWRPs making use of inline resources (there are 2 in the default recipe), we get more and more restarts of Splunk, which in turn will make our chef runs longer and longer.

Instead we probably need to make use of something like marker files... In the Splunk recipe, kick off a delayed notification to restart Splunk only if a marker file is present. In LWRPs (and elsewhere) if changes are made immediately touch the marker file, then when Splunk is successfully started with new config (or restarted) notify delete of the marker file. 

Does this make sense?

mattnelson:
Shouldn't one of the from the recipe be to `test_lwrp_notinline`?

https://github.com/acharlieh/test-lwrp/blob/master/recipes%2Fdefault.rb#L15-17

acharlieh:
This was quickly thrown together this afternoon as a POC. The point was to have both be the same. If we switch them to notinline and remove the bash declaration from within the default LWRP, thus mimicing current state, show that there's only one restart. I probably should have one recipe (and set of LWRPs) showing current state and one showing proposed state.

acharlieh:
Closing this PR per the above comment of the design adding restarts, we should pursue marker files as a solution.

PR #38, opened 2015-01-08
Fixes #3 - Fix splat operator in authorize.conf generation

### LINKED ISSUE #3: Overzealous use of Array Coersion
Apparently if a variable `x` is a `Hash`, Array coercion, e.g. `[*x]` results in an array of arrays instead of an array of a single Hash element. 

In the process of open sourcing and applying rubocop, I apparently overzealously replaced an instances of `[x] unless x.is_a?(Array)` where I was coercing a variable of an Array, Hash, or String to an Array.

https://github.com/cerner/cerner_splunk/blob/1.5.0/recipes/_configure_authentication.rb#L74

acharlieh:
This PR is for fixing Issue #3. 

There are two other places where the splat operator is in use, but in these cases, using the splat seems to be the correct behavior:
* [libraries/splunk_app.rb](https://github.com/cerner/cerner_splunk/blob/1.7.0/libraries/splunk_app.rb#L139) - role will either be a String or an Array 
* [vagrant_repo/cookbooks/cerner_splunk_test/templates/default/access.log.erb](https://github.com/cerner/cerner_splunk/blob/1.7.0/vagrant_repo/cookbooks/cerner_splunk_test/templates/default/access.log.erb) - This is an expansion of an explicit range in the test cookbook.

### DIFF
--- a/recipes/_configure_authentication.rb
+++ b/recipes/_configure_authentication.rb
@@ -71,7 +71,7 @@ when 'Scripted'
 when 'LDAP'
   strategies = hash.delete('LDAP_strategies')
   fail 'LDAP_strategies required for LDAP authentication' unless strategies
-  strategies = [*strategies]
+  strategies = [strategies] unless strategies.is_a? Array
   strategies = strategies.collect do |strategy|
     hash =
       case strategy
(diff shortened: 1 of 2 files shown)


david-crowder:
+1

aaronblythe:
+1

nimeshsubramanian:
+1

PR #40, opened 2015-01-13
Fixes #36 - Don't include empty apps in resultant merge

### LINKED ISSUE #36: The app data bag creates and empty "bag" directory
It appears as though the logic for dropping an app is a little overzealous and creates and empty "bag" application if that key is used in the apps data bag.

acharlieh:
For fixing #36 - I think this is a simple change, but we also introduce testing yay! :penguin: :bear:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,14 @@ source 'http://repo.release.cerner.corp/internal/rubygems/' if internal
 gem 'berkshelf', '~> 3.0', group: :deployment
 gem 'rubocop', '~> 0.18'
 gem 'foodcritic', '~> 4.0'
+gem 'rspec',  '~> 3.1'
+
+# https://github.com/opscode/chef/issues/2547
+if Bundler.current_ruby.on_19?
+  gem 'chef', '~> 11.0'
+else
+  gem 'chef', '~> 12.0'
+end
 
 group :development do
   gem 'roll_out', '~> 1.6.0'
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,4 @@ rvm:
 script:
 - bundle exec rubocop
 - bundle exec foodcritic -f any .
+- bundle exec rspec
--- a/libraries/splunk_app.rb
+++ b/libraries/splunk_app.rb
@@ -23,7 +23,7 @@ module CernerSplunk
           app_hash.merge!(to_merge, &merger)
         end
 
-        result[app_name] = app_hash
+        result[app_name] = app_hash unless app_hash.empty?
       end
     end
   end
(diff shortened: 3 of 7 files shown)


david-crowder:
+1

tests :tada:

Gemfile
```diff
@@ -10,6 +10,14 @@ source 'http://repo.release.cerner.corp/internal/rubygems/' if internal
 gem 'berkshelf', '~> 3.0', group: :deployment
 gem 'rubocop', '~> 0.18'
 gem 'foodcritic', '~> 4.0'
+gem 'rspec',  '~> 3.1'
+
+# https://github.com/opscode/chef/issues/2547
+if Bundler.current_ruby.on_19?
```
aaronblythe: Is this part of the context of this change?

I am concerned on this and would like to better understand:
* What version of ruby currently comes with ChefDK?
* Will this force users local version of Chef onto Chef 12 when they want to stay on Chef 11 in anyway?

acharlieh: It's a related change, as part of adding tests (since the file I was testing pulls in chef). I thought about splitting the file to not have the chef dependency but we'd still run into this sooner or later.

Chef 12 Omnibus (and ChefDK) comes with Ruby 2.1.4 but Chef 11 Omnibus ships with Ruby 1.9.3. [That's why we run our travis build on both|https://github.com/cerner/cerner_splunk/blob/1.7.0/.travis.yml#L7-L8]. 

When you are on ruby 1.9.3, adding `gem 'chef'` or `gem 'chef ~> 12'` as a dependency will fail to resolve with an install error namely with `ohai 8.0.1` because: `ohai requires Ruby version >= 2.0.0`. We could resolve Chef 11 on Ruby 2, but that somewhat defeats part of the reason for using Ruby 2 (which is what 12 comes with)

It could force local users using Ruby 2 to wind up with both gems installed, but ideally they're using gemsets or `bundle exec` to control what versions are resolved where. (Users on 1.9.3 could not upgrade to 12 due to the resolution problem)

aaronblythe: Should get interesting soon.  Ruby 1.9.3 end of development was a year ago.  End of security patches is one month away: https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/ So this is a larger discussion than this issue.

For this issue :shipit: but expect conversations on the change.

aaronblythe:
:+1:

acharlieh:
The merge commit is: 92db0df729f442da2d890bb2c88c17b5ff4fb813 (minor conflict merging both this branch and the process and internal tool update branch)

PR #44, opened 2015-02-06
Add sensitive attribute some resources

david-crowder:
Conditionally including the sensitive attribute to maintain backwards compatibility.

### DIFF
--- a/libraries/passive_sensitive.rb
+++ b/libraries/passive_sensitive.rb
@@ -1,0 +1,16 @@
+# coding: UTF-8
+#
+# Cookbook Name:: cerner_splunk
+# File Name:: passive_sensitive.rb
+
+require 'chef/resource'
+
+# rubocop:disable Documentation
+class Chef
+  # Making the sensitive attribute passive for older chef versions
+  class Resource
+    def sensitive(_ = nil)
+    end
+  end
+end unless Chef::Resource.method_defined? :sensitive
+# rubocop:enable Documentation
--- a/libraries/passive_sensitive.rb
+++ b/libraries/passive_sensitive.rb
@@ -9,7 +9,8 @@ require 'chef/resource'
 class Chef
   # Making the sensitive attribute passive for older chef versions
   class Resource
-    def sensitive(_ = nil)
+    def sensitive(args = nil)
+      set_or_return(:sensitive, args, kind_of: [TrueClass, FalseClass])
     end
   end
 end unless Chef::Resource.method_defined? :sensitive
--- a/libraries/splunk_template.rb
+++ b/libraries/splunk_template.rb
@@ -8,6 +8,7 @@ require 'chef/resource/template'
 require 'chef/provider/template'
 require 'chef/mixin/securable'
 require_relative 'lwrp'
+require_relative 'passive_sensitive'
 
 class Chef
   class Resource
--- a/recipes/_configure_authentication.rb
+++ b/recipes/_configure_authentication.rb
@@ -138,6 +138,7 @@ else
 end
 
 splunk_template 'system/authentication.conf' do
+  sensitive auth_stanzas.any? { |_, v| v.key? 'bindDNpassword' }
   stanzas auth_stanzas
   notifies :restart, 'service[splunk]'
 end
--- a/recipes/_generate_password.rb
+++ b/recipes/_generate_password.rb
@@ -17,6 +17,7 @@ new_password = SecureRandom.hex(36)
 
 execute 'change-admin-password' do
   command "#{node['splunk']['cmd']} edit user admin -password #{new_password} -roles admin -auth admin:#{old_password}"
+  sensitive true
 end
 
 if platform_family?('windows')
@@ -32,5 +33,6 @@ file password_file do
   owner system_user
   group system_group
   mode '0600'
+  sensitive true
   content new_password
 end
--- a/recipes/_generate_password.rb
+++ b/recipes/_generate_password.rb
@@ -15,7 +15,7 @@ password_file = File.join node['splunk']['external_config_directory'], 'password
 old_password = File.exist?(password_file) ? File.read(password_file) : 'changeme'
 new_password = SecureRandom.hex(36)
 
-execute 'change-admin-password' do
+execute 'change-admin-password' do # ~FC009
   command "#{node['splunk']['cmd']} edit user admin -password #{new_password} -roles admin -auth admin:#{old_password}"
   sensitive true
 end


acharlieh:
We'll get the new rubocop failure in a different PR. (TODO: link new PR for rubocop)

A few questions:
1) Should we use sensitive in the execute resource that's changing the admin password as well?
2) How about in the stanza writing authorize.conf that could (potentially) write an ldap bind password? (Maybe true if we're including the password, false if not?)
3) Would it be better to add a library file, that would monkeypatch in a single argument noop sensitive method (if it doesn't exist) so that we don't have to have the if check everywhere?

david-crowder:
1) Should we use sensitive in the execute resource that's changing the admin password as well?

Yes (though on [chef 11 versions it won't do too much](https://github.com/chef/chef/pull/2013)) 

2) How about in the stanza writing authorize.conf that could (potentially) write an ldap bind password? (Maybe true if we're including the password, false if not?)

Yes

3) Would it be better to add a library file, that would monkeypatch in a single argument noop sensitive method (if it doesn't exist) so that we don't have to have the if check everywhere?

I think it would be better to write a library if we are wanting to maintain chef 10 and early 11 compatibility (which I think we are).
I will make the suggested changes and update the PR :smile:

david-crowder:
Updated the PR, however the build will likely fail due to https://github.com/acrmp/foodcritic/issues/289

libraries/splunk_template.rb
```diff
@@ -60,6 +61,10 @@ def variables(args = nil)
         end
       end
 
+      def sensitive(args = nil)
```
acharlieh: Should our monkey patch just be this? Maybe with an additional `default: false` validator? This resource should then be able to use it passively without duplicating code (or overriding code in future chef versions).

libraries/passive_sensitive.rb
```diff
@@ -0,0 +1,13 @@
+# coding: UTF-8
+#
+# Cookbook Name:: cerner_splunk
+# File Name:: passive_sensitive.rb
+
+require 'chef/resource'
+
+class Chef
+  class Resource
+    def sensitive(arg)
```
acharlieh: Do we need the default argument for our monkey patch to be passive?

recipes/_configure_authentication.rb
```diff
@@ -138,6 +138,7 @@
 end
 
 splunk_template 'system/authentication.conf' do
+  sensitive true
```
acharlieh: How about `sensitive auth_stanzas.any? { |_, v| v.key? "bindDNpassword" }` since it's only sensitive if there's a bind password

libraries/passive_sensitive.rb
david-crowder: Rubocop is suggesting 
```
sensitive(*)
```

for the definition. We are only defining the method if it doesn't already exist. 
How would adding nil be more passive?

recipes/_configure_authentication.rb
david-crowder: +1

libraries/splunk_template.rb
david-crowder: Removing as it wasn't needed. 
Also, adding the require as I know we have run into issues with that in the past.

libraries/passive_sensitive.rb
acharlieh: The thought is that the [real definition](https://github.com/chef/chef/blob/12.0.3/lib/chef/resource.rb#L317) is `def sensitive(arg=nil)` (to satisfy rubocop we could do `def sensitive(_ = nil)` ) Which enables invoking both with and without a single argument. `def sensitive(*)` would enable passing more arguments than the real definition.

david-crowder: Updated per your recommendation. I kinda get why we want it to match the real definition. As I can come up with a few (some contrived) reasons for matching and can't come up with any reasons not to, I made the change :p

libraries/passive_sensitive.rb
```diff
@@ -0,0 +1,16 @@
...
+
+# rubocop:disable Documentation
+class Chef
+  # Making the sensitive attribute passive for older chef versions
+  class Resource
+    def sensitive(_ = nil)
+    end
+  end
+end unless Chef::Resource.method_defined? :sensitive
+# rubocop:enable Documentation
```
ncerny: Instead of making this passive, what would be the work to make it just perform the same as later versions of chef?  Would it just be: https://github.com/chef/chef/blob/647e90ab9b778e92777f54c7201f70656afcf6cc/lib/chef/provider/execute.rb#L89-L91 ?

recipes/_generate_password.rb
```diff
@@ -17,6 +17,7 @@
 
 execute 'change-admin-password' do
```
acharlieh: To get the build going could we disable the failing rule on this line [like we did in the provider](https://github.com/cerner/cerner_splunk/blob/1.7.0/providers/forwarder_monitors.rb#L8)?

libraries/passive_sensitive.rb
```diff
@@ -0,0 +1,16 @@
...
+# Cookbook Name:: cerner_splunk
+# File Name:: passive_sensitive.rb
+
+require 'chef/resource'
+
+# rubocop:disable Documentation
+class Chef
+  # Making the sensitive attribute passive for older chef versions
+  class Resource
+    def sensitive(_ = nil)
```
acharlieh: I'm going back and forth here, whether this should be a no-op (since noone uses this on earlier versions of chef), or if we should have this set and retrieve the property like the real version does, so we (and others) could potentially leverage it ourselves.

recipes/_generate_password.rb
david-crowder: +1, the PR to fix this issue has been sitting on the foodcritic project and I was hoping it would get merged soon. However, since we are ensuring backwards compatibility I am cool ignoring the rule on this resource.

libraries/passive_sensitive.rb
david-crowder: I don't think it would be prudent to duplicate the functionality of the sensitive flag for older versions of chef. There was a good amount of code that changed and I feel like maintenance would be a pain.

libraries/passive_sensitive.rb
david-crowder: I went ahead and added the functionality and pushed the new commit (I can squash later if you want). As far as I can see it won't produce any issues and if we want to consume it elsewhere we won't have to worry about breaking compatibility.

acharlieh:
:+1:

mccartjm:
+1

PR #48, opened 2015-03-03
Fixes #37 - Install apps via tarballs 

### LINKED ISSUE #37: Manage Tarballed Apps
Next we need to be able to download and install apps from tar.gz / spl files.

acharlieh:
Implementation of #37. This was quite the adventure. But this gives us the ability to install apps from .tar.gz files (and .spl, but .spl is just a renamed .tar.gz). We'll need a new enhancement / issue if we want to install apps from .zip files as well. (In the past few days I've seen a number zip files with all kinds of formats.)

Please review and be harsh! Question everything!

### DIFF
--- a/docs/databags.md
+++ b/docs/databags.md
@@ -93,7 +93,10 @@ An apps hash is a contextual (see above) Hash, part of a plaintext data bag item
 * `['bag']` - A string that points to an externalized Apps Hash in which all keys (except this one) are valid. 
 * `[app]` - The name of an app to manage (disk name)
 * `[app]['remove']` - If true, remove this app instead of creating / managing  (default - false)
-* `[app]['local']` - If true, manage files in the local directories instead of the "default" (default-false)
+* `[app]['local']` - If true, manage `[app]['files']` defined files and `[app]['permissions']` defined metadata in the "local" directory and "local.meta" instead of the "default" directory and "default.meta" (default - false, but forced true if download-url is specified)
+* `[app]['download']` - Information for downloading an app
+* `[app]['download']['url']` - URL of where to download the app .tar.gz or .spl file. Archive is expected to contain a top-level directory with name matching 'app' attribute above.
+* `[app]['download']['version']` - Expected [version number][app.conf] (if any) used to determine if a new app should be downloaded.
 * `[app]['files']` - Hash of files to manage under the "default" or "local" directory.
 * `[app]['files'][filename]` - Contents of a particular file to manage. It can take 3 values, a hash of stanzas -> key-value pairs (then written with the splunk template), a string (written as is), or nil / false (deleted). If the hash or string is empty, the file is also deleted.
 * `[app]['permissions']` - Hash of permissions to manage for the app.
@@ -115,3 +118,4 @@ Docs Navigation
 [authentication.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Authenticationconf
 [alert-actions.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Alert-actionsconf
 [default.meta]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Defaultmetaconf
+[app.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/appconf
--- a/libraries/conf.rb
+++ b/libraries/conf.rb
@@ -4,7 +4,39 @@
 # File Name:: conf.rb
 
 module CernerSplunk
+  # Utilities for working with splunk configuration files
   module Conf
+    STANZA_START ||= /^\s*\[(?<stanza_name>[^\]]+)\].*$/
+    COMMENT_LINE ||= /^\s*#.*$/
+    KEY_VALUE_PAIR ||= /^(?<key>[^=]+)=(?<value>.*)$/
+
+    def self.parse_string(conf)
+      return {} if conf.nil?
+      parse StringIO.open(conf, 'r:UTF-8')
+    end
+
+    def self.parse(io)
+      start = {}
+      hash = { 'default' => start }
+      current = start
+
+      io.each_line do |line|
+        case line
+        when STANZA_START
+          start = {}
+          current = start
+          hash[Regexp.last_match(:stanza_name).strip] = start
+        when COMMENT_LINE
+          # ignore comment lines
+        when KEY_VALUE_PAIR
+          current[Regexp.last_match(:key).strip] = Regexp.last_match(:value).strip
+        end
+      end
+
+      hash.delete('default') if hash['default'].empty?
+      hash
+    end
+
     # Class for reading an existing configuration file from disk
     class Reader
       def initialize(filename)
@@ -13,27 +45,10 @@ module CernerSplunk
 
       def read
         return {} unless File.exist?(@filename)
-
-        start = {}
-        hash = { 'default' => start }
-        current = start
-
-        File.open(@filename, 'r:UTF-8').each_line do |line|
-          case line
-          when /^\s*\[([^\]]+)\].*$/
-            start = {}
-            current = start
-            hash[Regexp.last_match[1].strip] = start
-          when /^\s*#.*$/
-            # ignore comment lines
-          when /^([^=]+)=(.*)$/
-            current[Regexp.last_match[1].strip] = Regexp.last_match[2].strip
-          end
-        end
-
-        hash.keep_if do |_, value|
-          !value.empty?
-        end
+        file = File.open(@filename, 'r:UTF-8')
+        CernerSplunk::Conf.parse file
+      ensure
+        file.close if file
       end
     end
   end
--- a/libraries/splunk_app.rb
+++ b/libraries/splunk_app.rb
@@ -27,6 +27,22 @@ module CernerSplunk
       end
     end
   end
+
+  # Utility Class to parse app version strings
+  class AppVersion
+    def initialize(version)
+      @version = version.chomp.empty? ? nil : version.chomp unless version.nil?
+      @base, @prerelease = @version.split(' ', 2) unless @version.nil?
+      @type = @prerelease.nil? ? :base : :prerelease unless @version.nil?
+    end
+
+    attr_reader :version, :base, :prerelease, :type
+    alias_method :to_s, :version
+
+    def ==(other)
+      version == other.version
+    end
+  end
 end
 
 class Chef
--- a/libraries/splunk_app.rb
+++ b/libraries/splunk_app.rb
@@ -81,6 +81,14 @@ class Chef
         set_or_return(:app, arg, kind_of: String)
       end
 
+      def url(arg = nil)
+        set_or_return(:url, arg, kind_of: String)
+      end
+
+      def version(arg = nil)
+        set_or_return(:version, arg, kind_of: String)
+      end
+
       # Calculated attributes
       def required_directories
         %w(local default metadata lookups).collect { |d| "#{root_dir}/#{d}" }.unshift(root_dir)
@@ -106,10 +114,12 @@ class Chef
   end
 end
 
+require_relative 'conf'
+
 class Chef
   class Provider
     # Chef Provider for managing Splunk apps
-    class SplunkApp < Chef::Provider
+    class SplunkApp < Chef::Provider # rubocop:disable ClassLength
       def whyrun_supported?
         true
       end
@@ -126,7 +136,8 @@ class Chef
       end
 
       def action_create
-        create_app_directories
+        download_and_install
+        create_app_directories unless new_resource.updated_by_last_action?
         manage_metaconf unless new_resource.permissions.empty?
         new_resource.files.each do |file_name, contents|
           *directories, file_name = file_name.split('/')
@@ -164,6 +175,91 @@ class Chef
         dir.mode('0755')
         dir.run_action(:create)
         new_resource.updated_by_last_action(true) if dir.updated_by_last_action?
+      end
+
+      def should_download?(expected_version, installed_version)
+        # No need to download if we already have the exact version installed
+        return false if expected_version.version && expected_version == installed_version
+        # We must not install a prerelease on top of the same released version
+        fail "Expecting to install prerelease on top of same released version for #{new_resource.app}" if installed_version.type == :base && expected_version.base == installed_version.base
+        # Warn (but not fail) if the expected version is not specified. (Optimization for us)
+        Chef::Log.warn "Expected version not specified for #{new_resource.app}." unless expected_version.version
+        # When in whyrun mode, we want to stop here as the rest requires the tarball to be downloaded (thus changing the node).
+        !whyrun_mode?
+      end
+
+      def download_and_install
+        return if new_resource.url.nil? || new_resource.url.empty?
+
+        expected_version = CernerSplunk::AppVersion.new new_resource.version
+        installed_version = CernerSplunk::AppVersion.new @current_resource.version
+
+        return unless should_download? expected_version, installed_version
+
+        filename = "#{Chef::Config[:file_cache_path]}/#{new_resource.app}.tgz"
+
+        download = Chef::Resource::RemoteFile.new(filename, run_context)
+        download.source(new_resource.url)
+        download.run_action(:create)
+        download.backup(false)
+
+        install_from_tar filename, expected_version, installed_version
+      ensure
+        download.run_action(:delete) if download
+      end
+
+      def validate_downloaded(tarfile)
+        fail "Downloaded tarball from '#{new_resource.url}' does not contain an app named '#{new_resource.app}'" if tarfile.num_files == 0
+        fail "Downloaded tarball for '#{new_resource.app}' has local entries" unless tarfile.count { |p, _| p.match %r{^[^/]+/local/.+} } == 0
+      end
+
+      def should_install?(expected_version, installed_version, tar_version) # rubocop:disable PerceivedComplexity, CyclomaticComplexity
+        fail "Downloaded tarball for #{new_resource.app} does not contain a version in app.conf!" unless tar_version.version
+        # If we specify an expected version (see warning in should download), the tar version must match exactly OR the expected version is the base version of the (prerelease) tar version
+        if expected_version.version && tar_version != expected_version
+          fail "Expected version #{expected_version} does not match tar version #{tar_version} for #{new_resource.app}" unless expected_version.type == :base && tar_version.base == expected_version.base
+        end
+        # If the exact version is already installed, NOOP
+        return false if tar_version == installed_version
+        # We must not install a prerelease on top of the same released version
+        fail "Attempting to install prerelease on top of same released version for #{new_resource.app}" if installed_version.type == :base && installed_version.base == tar_version.base
+        true
+      end
+
+      def install_from_tar(filename, expected_version, installed_version)
+        tarfile = CernerSplunk::TarBall.new(filename, prefix: new_resource.app, user: node['splunk']['user'], group: node['splunk']['group'])
+
+        validate_downloaded tarfile
+
+        app_conf = CernerSplunk::Conf.parse_string tarfile.get_file('default/app.conf')
+        tar_version = CernerSplunk::AppVersion.new((app_conf['launcher'] || {})['version'])
+
+        return unless should_install? expected_version, installed_version, tar_version
+
+        old_dir_path = "#{new_resource.root_dir}.old"
+
+        old_dir = Chef::Resource::Directory.new(old_dir_path, run_context)
+        old_dir.recursive true
+        old_dir.run_action :delete
+
+        # Move existing app out of the way
+        ::File.rename new_resource.root_dir, old_dir_path if ::File.exist? new_resource.root_dir
+        # Extract tarball to app directory
+        tarfile.extract new_resource.apps_dir
+
+        # Restore all potential user defined content
+        create_app_directories
+        ::Dir.chdir old_dir_path do
+          ::Dir['local/*', 'lookups/*', 'metadata/local.meta'].each do |f|
+            ::File.rename "#{old_dir_path}/#{f}", "#{new_resource.root_dir}/#{f}"
+          end
+        end if ::File.exist? old_dir_path
+        # Remove old app
+        old_dir.run_action :delete
+
+        new_resource.updated_by_last_action true
+      ensure
+        tarfile.close if tarfile
       end
 
       def manage_metaconf
--- a/libraries/splunk_app.rb
+++ b/libraries/splunk_app.rb
@@ -49,7 +49,8 @@ class Chef
       end
 
       def local(arg = nil)
-        set_or_return(:local, arg, kind_of: [TrueClass, FalseClass])
+        val = set_or_return(:local, arg, kind_of: [TrueClass, FalseClass])
+        (url.nil? || url.empty?) ? val : true
       end
 
       def files(arg = nil)
@@ -62,6 +63,28 @@ class Chef
 
       def app(arg = nil)
         set_or_return(:app, arg, kind_of: String)
+      end
+
+      # Calculated attributes
+      def required_directories
+        %w(local default metadata lookups).collect { |d| "#{root_dir}/#{d}" }.unshift(root_dir)
+      end
+
+      def root_dir
+        "#{apps_dir}/#{app}"
+      end
+
+      def default_dir
+        "#{root_dir}/default"
+      end
+
+      def files_dir
+        local ? "#{root_dir}/local" : default_dir
+      end
+
+      def perms_file
+        file_name = local ? 'local.meta' : 'default.meta'
+        "#{root_dir}/metadata/#{file_name}"
       end
     end
   end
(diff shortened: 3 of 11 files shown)


gseats:
+1

vagrant_repo/apps/my_app/default/data/ui/nav/default.xml
```diff
@@ -0,0 +1,3 @@
+<nav search_view="search" color="#72231F">
+  <view name="lorem_ipsum" default="true" />
+</nav>
```
gseats: No newline at end of file.

acharlieh: Fixed with a44768b5e83fabfe91ba8c0859698fab79b9f73b (being the top commit I rewrote history)

libraries/conf.rb
```diff
@@ -4,7 +4,35 @@
...
+      io.each_line do |line|
+        case line
+        when /^\s*\[([^\]]+)\].*$/
+          start = {}
+          current = start
+          hash[Regexp.last_match[1].strip] = start
+        when /^\s*#.*$/
+          # ignore comment lines
+        when /^([^=]+)=(.*)$/
+          current[Regexp.last_match[1].strip] = Regexp.last_match[2].strip
```
garrypolley: Maybe I'm splitting hairs but I'd prefer if these regexes where defined as a compiled constant and then used instead of relying on the state of `Regex.last_match`. 

Hopefully that makes sense.

libraries/splunk_app.rb
```diff
@@ -126,13 +174,80 @@ def create_splunk_directory(path)
...
+        return if new_resource.url.nil? || new_resource.url.empty?
+
+        expected_version = CernerSplunk::AppVersion.new new_resource.version
+        installed_version = CernerSplunk::AppVersion.new @current_resource.version
+
+        return if expected_version.type && expected_version == installed_version
+        Chef::Log.warn "Expected version not specified for #{new_resource.app}." unless expected_version.type
+        fail "Expecting to install prerelease on top of same released version for #{new_resource.app}" if expected_version.type == :prerelease && installed_version.type == :base && expected_version.base == installed_version.base
+
+        # In whyrun mode, the rest is not safe.
```
garrypolley: Why is it not safe?

libraries/splunk_app.rb
```diff
@@ -126,13 +174,80 @@ def create_splunk_directory(path)
...
+        tarfile = CernerSplunk::TarBall.new(filename, prefix: new_resource.app, user: node['splunk']['user'], group: node['splunk']['group'])
+
+        fail "Downloaded tarball from '#{new_resource.url}' does not contain an app named '#{new_resource.app}'" if tarfile.num_files == 0
+
+        fail "Downloaded tarball for '#{new_resource.app}' has local entries" unless tarfile.count { |p, _| p.match %r{^[^/]+/local/.+} } == 0
+
+        app_conf = CernerSplunk::Conf.parse_string tarfile.get_file('default/app.conf')
+        tar_version = CernerSplunk::AppVersion.new((app_conf['launcher'] || {})['version'])
+
+        fail "Bad Developer!!!! The downloaded tar for #{new_resource.app} does not contain a version in app.conf!" if tar_version.type.nil?
```
garrypolley: Message could have a kinder tone.

libraries/splunk_app.rb
acharlieh: Because from this point forward we start altering the node (beginning with downloading the tarball)

libraries/splunk_app.rb
```diff
@@ -126,13 +174,80 @@ def create_splunk_directory(path)
...
+        download.source(new_resource.url)
+        download.run_action(:create)
+        download.backup(false)
+
+        install_from_tar filename, expected_version, installed_version
+      ensure
+        download.run_action(:delete) if download
+      end
+
+      def install_from_tar(filename, expected_version, installed_version) # rubocop:disable PerceivedComplexity, CyclomaticComplexity
```
garrypolley: This is doing a lot.  Is there a way to split out some of the reading into helper functions? 

Like move the untaring to it's own function and the validation into it's own function? Just thoughts to maybe make this a little easier for humans to parse.  At least this human,  :smiley_cat: .

libraries/tarball.rb
```diff
@@ -0,0 +1,141 @@
...
+        FileUtils.chown @options[:user], @options[:group], target
+      end
+
+      iterate file: file_handler, directory: dir_handler, symlink: symlink_handler
+    end
+
+    def count(callbacks = {}, &other)
+      callbacks = { other: other } if other && callbacks.empty?
+      i = 0
+      hsh = callbacks.each_with_object({}) { |(k, v), h| h[k] = proc { |*args| i += 1 if v.call(args) } }
```
garrypolley: some doc here would help me out. I'm not sure what's going on with the callbacks and how it relates to this count that actually gets returned.

libraries/tarball.rb
```diff
@@ -0,0 +1,141 @@
...
+      segments = path.split('/')
+      segments = segments[0...-1] unless all
+      segments.each_with_object(extract_dir.dup) do |segment, target|
+        target << "/#{segment}"
+        FileUtils.mkdir target unless File.directory? target
+        FileUtils.chown @options[:user], @options[:group], target
+      end
+    end
+
+    def iterate(callbacks = {}) # rubocop:disable PerceivedComplexity, CyclomaticComplexity
```
garrypolley: In my ideal world there would be a lot more inline doc explaining why some things are happening in here.  I usually prefer a comment as to why the cop is disabled in most Ruby code.

vagrant_repo/apps/my_app/default/data/ui/views/lorem_ipsum.xml
```diff
@@ -0,0 +1,10 @@
+<dashboard>
+  <label>Lorem Ipsum</label>
+  <description>Hurray!</description>
+  <row>
+    <html>
+      <img src="http://33.media.tumblr.com/tumblr_lzkpw7DAl21qhy6c9o4_250.gif"/>
```
garrypolley: :smile:   fancy

libraries/splunk_app.rb
garrypolley: That makes sense.  Do you mind adding a commit to that effect here?  Something like: 

```rb
# When in whyrun mode, stop so we do not modify the node by adding a tarball.
```

libraries/conf.rb
acharlieh: Hm, note to self, don't link from commits as I'm going to squash everything at the end. But is this what you had in mind: 34a4d063b35f41a9e65c3de1bf5a98fa9a8f8e2a

garrypolley: Yep that looks good to me.

vagrant_repo/apps/my_app/default/data/ui/views/lorem_ipsum.xml
acharlieh: Indeed. Getting toward the end, I felt I needed some festivity. I could have done 강남스타일

david-crowder: I think that is the best gif of a cat I have ever seen

libraries/splunk_app.rb
```diff
@@ -27,6 +27,22 @@ def self.merge_hashes(*hashes)
       end
     end
   end
+
+  # Utility Class to parse app version strings
+  class AppVersion
+    def initialize(version)
+      @version = version.chomp.empty? ? nil : version.chomp unless version.nil?
+      @base, @prerelease = @version.split(' ', 2) unless @version.nil?
+      @type = @prerelease.nil? ? :base : :prerelease unless @version.nil?
```
david-crowder: Couldn't you wrap all these in a single unless version.nil?

acharlieh: I could wrap the last two lines in an `unless @version.nil?` The first line is out because if `version` is only whitespace or the empty string `@version` is `nil`. 

But when I then do so, I wind up with a rubocop warning of `Use a guard clause instead of wrapping the code inside a conditional expression`

libraries/splunk_app.rb
acharlieh: Proposed: db6d88858e152daca29bf078ab7f80d1e0548405

    # When in whyrun mode, we want to stop here as the rest requires the tarball to be downloaded (thus changing the node).

libraries/splunk_app.rb
acharlieh: Proposed: db6d88858e152daca29bf078ab7f80d1e0548405

Removed the Bad Developer part. If the output was visual though...

nope badger

:grinning:

garrypolley: Change looks good to me.  I support gif.

libraries/splunk_app.rb
garrypolley: That's good.  Let's me know why it's there.

libraries/splunk_app.rb
david-crowder: oh, good to know :smile:

david-crowder:
+1

libraries/splunk_app.rb
acharlieh: Is caf34c9ef5750377f7d50d0e1299556daaf23779 better, or would you like more? 

Broke out the validations and added comments to increase readability.

garrypolley: Looks good to me.  My human brain can now understand it.

docs/databags.md
```diff
@@ -93,7 +93,10 @@ An apps hash is a contextual (see above) Hash, part of a plaintext data bag item
 * `['bag']` - A string that points to an externalized Apps Hash in which all keys (except this one) are valid. 
 * `[app]` - The name of an app to manage (disk name)
```
ben-roling: As I was reading this and looking at [the example below](https://github.com/cerner/cerner_splunk/pull/48/files#diff-cb44ff154fe5f501a6e91864def7eccaR4), I got to wondering about the expected structure of the tarball.  The way I have seen app tarballs (and the way our team built our own tarball) is with a top-level directory that has the app name but since you have a name here it is somehwhat ambiguous as to whether the tarball is expected to have that top-level directory layer.  If I'm reading the code correctly you do expect that layer to be there but it would be nice to have that made a bit more explicit in the doc.  Something else related would be question as to what happens if the name of that top-level directory differs from the name specified here?

docs/databags.md
```diff
@@ -93,7 +93,10 @@ An apps hash is a contextual (see above) Hash, part of a plaintext data bag item
...
 * `[app]['remove']` - If true, remove this app instead of creating / managing  (default - false)
-* `[app]['local']` - If true, manage files in the local directories instead of the "default" (default-false)
+* `[app]['local']` - If true, manage files in the local directories instead of the "default" (default - false, but forced true if download-url is specified)
```
ben-roling: The description of this attribute doesn't really make sense to me.  Maybe you can explain a bit further?

docs/databags.md
acharlieh: The expected structure of the tarball is indeed that there is a top level directory that is the app name. (Since this databag is used to configure all apps, including those generated by chef, I need to know the name of the tarball). The example app in vagrant_repo/apps/my_app is built in this way by [in the Vagrantfile above](https://github.com/acharlieh/cerner_splunk/blob/a44768b5e83fabfe91ba8c0859698fab79b9f73b/Vagrantfile#L158)

If the tarball does not contain the named app, (as would be the case when the names don't match), the chef run [fails in the splunk_app resource execution, below](https://github.com/acharlieh/cerner_splunk/blob/a44768b5e83fabfe91ba8c0859698fab79b9f73b/libraries/splunk_app.rb#L208)

Another note. If there are multiple top level directories in a particular tarball, we only install the one that we want to install (per the configuration).

ben-roling: Thanks.  I get it all now.

Maybe worth a slight amendment to [the description of the 'url' attribute](https://github.com/cerner/cerner_splunk/pull/48/files#diff-a8cd9604d38ad43b8b92ae521a075a6aR98) to say something like the following?

"URL of where to download the app .tar.gz or .spl file.  Archive is expected to contain top-level directory with name matching 'app' attribute above."

docs/databags.md
acharlieh: Splunk apps have two different folders for configuration. app_folder/default/ and app_folder/local/ The intent being that apps are shipped with "default" configuration, and people installing the apps (and those using the webui) can override / modify configurations though the local directory. (This precedence is described in the [splunk docs](http://docs.splunk.com/Documentation/Splunk/latest/Admin/Wheretofindtheconfigurationfiles) ) 

When this is being used by Chef to generate an app from scratch, we would want to write our configurations to default (as people using the webui would want to modify things freely). When we are modifying configuration for an app from a tarball, or an app that's pre-installed with the core Splunk package, any customizations we manage from Chef we would want to write to the local directory (the default directory being what is shipped with the app itself). 

We can't distinguish if we're working with a pre-installed app or an app generated by Chef (since on subsequent chef-client runs they would look identical), however an app that we install via Tarball, we would know that we must use the local directory.

docs/databags.md
acharlieh: I like it. Added to my PR changes branch (which will be squash merged after everything): 0c767d53791f107ca83f0a826f9b22bd67a7c551

libraries/tarball.rb
acharlieh: I have a first pass at it 63280a41fabbf0b60c05de65b7051885595b6504. 

But basically, callbacks are invoked as we iterate over the tarball. On this particular line, we are wrapping every Proc passed to the method, such that when the proc is invoked, if the inner proc returns a truthy value, we increment our counter.

libraries/tarball.rb
acharlieh: I have a first pass here: 63280a41fabbf0b60c05de65b7051885595b6504

Does this make more sense? Anything in particular that seems odd?

docs/databags.md
ben-roling: I think maybe I get it now.  When I was originally reading I was interpreting the attribute as having some effect on how the contents of the tarball itself would be treated.  It made me think as though the contents of the tarball are going to be treated as "local" even though typically what you'd put in a tarball would be "default".

I suspect this particular property applies the files described in [app]['files'] and whether those are treated as "local" or "default".  Naturally if you are installing a tarball and additionally defining [app]['files'] then what's in the tarball should be "default" and what is in [app]['files'] should be considered "local".

Maybe the change I can propose is to say the following:

If true, manage [app]['files'] defined files in the "local" directories instead of "default"...

The key difference there is I replaced the word "files" with "[app]['files']" to clarify which files exactly are being referred to.

acharlieh: Updated ef935041bd357ccbb7a8a507d7b36e2f47d58269

libraries/splunk_app.rb
```diff
@@ -126,13 +174,80 @@ def create_splunk_directory(path)
...
+        new_resource.updated_by_last_action(true) if dir.updated_by_last_action?
+      end
+
+      def download_and_install # rubocop:disable PerceivedComplexity, CyclomaticComplexity
+        return if new_resource.url.nil? || new_resource.url.empty?
+
+        expected_version = CernerSplunk::AppVersion.new new_resource.version
+        installed_version = CernerSplunk::AppVersion.new @current_resource.version
+
+        return if expected_version.type && expected_version == installed_version
```
ben-roling: I don't understand the first part of this expression (`if expected_verison.type`).  `expected_version.type` is not a boolean -- if I read [AppVersion](https://github.com/cerner/cerner_splunk/pull/48/files#diff-940247b1e7418e1207338ac1de247744R36) correctly the 'type' is either `:base` or `:prerelease` and never null?

libraries/splunk_app.rb
```diff
@@ -126,13 +174,80 @@ def create_splunk_directory(path)
...
+        Chef::Log.warn "Expected version not specified for #{new_resource.app}." unless expected_version.type
```
ben-roling: I'm confused by the `unless expected_version.type` here as well for the same reason as my confusion on the previous line.

libraries/splunk_app.rb
acharlieh: AppVersion#type will be nil if the version string given to the constructor is nil or blank. (the unless clause on the line you reference).

libraries/splunk_app.rb
acharlieh: Same explanation [as the previous line](https://github.com/cerner/cerner_splunk/pull/48/#discussion_r25885375). A version was not specified in the attributes / data bag.

libraries/splunk_app.rb
ben-roling: Ok, then maybe it would make more sense to check AppVersion#version instead of AppVersion#type?  At least that would look more straightforward to me.

ben-roling: Something else I'm a bit confused about on this line.  It looks like you have [defined ==](https://github.com/cerner/cerner_splunk/pull/48/files#diff-940247b1e7418e1207338ac1de247744R43) as only comparing AppVersion#version so if, for example, I installed "1.0 beta1" and then tried to update to "1.0 beta2", it would result in a no-op?

acharlieh: AppVersion#version is the whole string. (e.g. "1.0 beta1" or "1.0 beta2"). That's why == uses version for comparison. AppVersion#base would be "1.0" in both of those cases.

ben-roling: Sorry, I think you can disregard my previous comment -- I had myself confused about what was in 'version' -- I had it in my head that was the 'base' part of the version only.

libraries/splunk_app.rb
```diff
@@ -126,13 +174,80 @@ def create_splunk_directory(path)
...
+        app_conf = CernerSplunk::Conf.parse_string tarfile.get_file('default/app.conf')
+        tar_version = CernerSplunk::AppVersion.new((app_conf['launcher'] || {})['version'])
+
+        fail "Bad Developer!!!! The downloaded tar for #{new_resource.app} does not contain a version in app.conf!" if tar_version.type.nil?
+
+        if expected_version.type && tar_version != expected_version
+          fail "Expected version #{expected_version} does not match tar version #{tar_version} for #{new_resource.app}" unless expected_version.type == :base && tar_version.base == expected_version.base
+        end
+
+        return if tar_version == installed_version
```
ben-roling: I noticed in your test you have my_app as [1.0 SNAPSHOT in the app.conf](https://github.com/cerner/cerner_splunk/pull/48/files#diff-9cee7098655ddcc82045b71f38745918R2) but then [replace "SNAPSHOT" with a timestamp during provisioning of the VM](https://github.com/cerner/cerner_splunk/pull/48/files#diff-23b6f443c01ea2efcb4f36eedfea9089R156).  I guess if you didn't do something like that then this line would prevent newer versions of the SNAPSHOT from being installed on subsequent chef-client runs with a newer 1.0-SNAPSHOT tar, right?

My team has an app tarball but currently our snapshot tarball builds would always have something like "1.0 SNAPSHOT" in the app.conf.  Actually, we currently have "1.0-SNAPSHOT" but reviewing this I noticed that is a violation of the app.conf spec as it dictates a space between the base an pre-release version portions.  Anyway, it seems we will need to do some replacement of "SNAPSHOT" with a timestamp or build number for this to work for us.  That should be ok, but one other thing we are doing is setting the build number in the app.conf and we are setting that differently on every build.  An alternative would be for you to look at that as well.  Then perhaps you would re-deploy if the build numbers differ even though the versions match.  I do sort of like putting a build identifier in the pre-release portion of the version itself though as it makes the exact SNAPSHOT version visible in installed apps list in the Web UI.

I guess the question is what are your thoughts about the build attribute?  The spec suggests it is required and you have left it off you my_app app.conf example BTW.

libraries/tarball.rb
```diff
@@ -0,0 +1,141 @@
...
+# Methods for working with data from a tarball.
+
+require 'rubygems/package'
+require 'zlib'
+require 'fileutils'
+
+module CernerSplunk
+  # Utility class to encapsulate operations on tar.gz files
+  # Mad Props to http://stackoverflow.com/a/19139114/504685
+  class TarBall
```
ben-roling: I think I have to just admit defeat in terms of reviewing this class.  Parts of my trouble are probably my limited experience and skills in Ruby and parts are probably the overall complexity of the code in general.  Anyhow, I just want to note that I'm not really reviewing this specific piece.  I suspect others at least more versed in Ruby will have looked over it.

libraries/splunk_app.rb
acharlieh: Originally you're correct, I was considering using the build number as a determinator as well. But then I was reading [the app.conf spec](http://docs.splunk.com/Documentation/Splunk/6.2.1/admin/appconf) I noticed that the comments on build include "Increment this whenever you change files in appserver/static." and "Ensures browsers don't use cached copies of old static files in new versions of your app." which made me realize that it has more to do with controlling the browser cache than anything else (and doesn't really apply for apps without UI). Surveying some apps from the community site, it seems rather hit and miss as to if the build attribute is present at all, but they universally seem to have the version attribute. Additionally, the version attribute is displayed in the App manager UI, but it does not show build. (The corresponding REST API returns build number `| rest /services/apps/local | fields build, version, title, splunk_server` )

By limiting the selector to the single version attribute, we also simplify the decision of do we reinstall or not as well. (all of the expected vs tar vs actual logic that exists with version duplicated with build as well).

ben-roling: Yeah, the additional complexity of looking at the build number is probably not necessary.   I'm ok with the logic you have.  I think we can change our snapshot tarball builds to have a more dynamic pre-release version number component in the app.conf and call that good.

libraries/splunk_app.rb
acharlieh: Here's it using AppVersion#version instead of AppVersion#type: b3dfa6c81238f25ae9d616719e0eded55ba7e8bd

acharlieh:
Thanks everyone for your comments thus far!

I'm going to apply this rebase, and force push to update this PR:
 
    pick c484410 Documentation around apps from tarballs
    f 0c767d5 Update to download url attribute documentation
    f ef93504 Update to local documentation description
    pick 617d4e7 Updates to splunk_app RP 
    pick e77ed8a Utility updates 
    f 34a4d06 Updating conf.rb regexp handling
    f 63280a4 First pass at descriptions on how count and iterate work
    pick 8ce18cd Implement managing of an app from a remote tarball
    f db6d888 Updating splunk_app messaging for why run and no version.
    f caf34c9 Helper methods and comments in splunk_app for readability
    f b3dfa6c Use nil version instead of nil type checks in version comparisons.
    pick a44768b Include an app installed via tarball in Vagrant boxes

acharlieh:
Fixed trailing whitespace derp from 63280a41fabbf0b60c05de65b7051885595b6504 that wound up in 858cab699be902ac6f28b92db373276efb85eb5a in 43ebd3fbb23de3c4e208083154a5540225a433b6

vagrant_repo/apps/my_app/default/data/ui/views/lorem_ipsum.xml
```diff
@@ -0,0 +1,10 @@
...
+      <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam blandit fringilla ex, nec vehicula nunc ornare efficitur. Maecenas auctor lacus justo, imperdiet dapibus eros pharetra sed. Proin eget erat varius, elementum lorem non, aliquet ante. Fusce dapibus ligula sit amet velit blandit, facilisis fringilla erat ornare. Sed id neque in augue maximus mattis a quis orci. Quisque consequat nibh nulla, a sagittis eros imperdiet sed. Sed at risus a nisi placerat egestas. Vivamus aliquet venenatis lacus, molestie tempor odio gravida posuere. Donec eros lectus, gravida id placerat eu, rhoncus eu metus. Donec eget nisl lobortis, fringilla massa non, accumsan purus. Integer eget orci consequat, efficitur sapien eget, ornare odio. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Aenean pretium fringilla mauris vitae vulputate. Duis aliquam tellus vel ex imperdiet gravida. Donec tempus consequat volutpat.</p>
```
gseats: I was so confused for the longest time over this, then i googled the label.  -_-

garrypolley:
:+1:   Looks fine to me.

ben-roling:
+1

libraries/tarball.rb
```diff
@@ -0,0 +1,157 @@
...
+      segments = segments[0...-1] unless all
+      segments.each_with_object(extract_dir.dup) do |segment, target|
+        target << "/#{segment}"
+        FileUtils.mkdir target unless File.directory? target
+        FileUtils.chown @options[:user], @options[:group], target
+      end
+    end
+
+    # Iterate over the entries of the tarfile, and invoke the corresponding callback with each entry.
+    # We handle the LongFileName and LongSymLink entries as found (no callbacks, but they change values sent with other callbacks)
```
aaronblythe: I believe I follow most of what is going on here from reading the comments and:
* http://stackoverflow.com/questions/856891/unzip-zip-tar-tag-gz-files-with-ruby
* http://stackoverflow.com/questions/2078778/what-exactly-is-the-gnu-tar-longlink-trick

Is there anyway that you can provide a "why?" all of this indirection is needed, that is a little less technical.  Forgive me if this is here and I am simply missing it.

acharlieh: Which indirection(s) specifically are you questioning?

1. I put this library together, to eliminate declaring a dependency on a native tar package being present (and avoid building indirection to work natively on both Windows and *nix systems.)
1. LongFileName (L) and LongSymLink (K) are separate entries in GNU tar files that exist if the subsequent "file" (file here being in terms of *nix everything is a file, so could be a file, symlink, directory, or something else) path or symlink target are over 99 ASCII characters long (100 characters with null). POSIX header struct: http://www.gnu.org/software/tar/manual/html_node/Standard.html
1. The indirection of including only those with a certain prefix is so that we could extract a single app if the tarfile contains multiple apps. Also gives us an early catch of the tarfile doesn't contain the app you think it does.
1. The indirection of providing callbacks is similar to the Visitor / Iterator pattern, where I can declare each operation (extract, find file, count) simply in terms of what do I do when I encounter a particular type of entry (:file, :directory, :symlink) or any other entry (:other) instead of reimplementing the same iteration and prefix filter logic over and over again.

acharlieh: First pass at some more doc on the tarball class. Does this help? bb7eb4c2e7c40f6457b9651d74e1280aab142a65

nimeshsubramanian:
+1

Issue #57, opened 2015-04-01
Add Code Climate Test Coverage


acharlieh:
See:
http://docs.travis-ci.com/user/code-climate/#Measuring-Test-Coverage-with-Code-Climate
http://docs.codeclimate.com/article/219-setting-up-test-coverage

PR #61, opened 2015-04-28
Code climate badge for code score

garrypolley:
This is an interesting metric and I think it could help keep the code in a better state over time. It exists on some other repos and I figured we may as well try it here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@ Installs and Configures Splunk forwarders and servers, and other components rela
 
 Based on the work done by [BBY Solutions](https://github.com/bestbuycom/splunk_cookbook) and a previous Cerner team.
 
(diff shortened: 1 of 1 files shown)


acharlieh:
:+1: I think we should consider fixing #57 and show how terrible test coverage there currently is on this thing too. While we're at it we could also consider adding other http://shields.io/ as well (version from supermarket might be useful)

garrypolley:
It'd be good to add the released version that is out as well.  I'm not seeing a simple way to add a released version that does not involve manual updates.  I'd prefer to have a service or something that did it automatically.

garrypolley:
I figured out how to get the version on the readme.  Badges seem to be on the rise: 

http://badge.fury.io/for/gh/cerner/cerner_splunk

There is actually an entire repo dedicated to proper badges: 

https://github.com/badges/shields

Which is the site you linked, and it has other sets of sites linked there.

acharlieh:
Added the webhook as mentioned in the badge fury site. Going to merge with the green button since it's a doc only change.

PR #62, opened 2015-05-26
Added configuration option for limits.conf

bbrownz:
We've ran across a couple situations where we've been forwarding a large amount of data from a single node. This adds the capability to configure the limits.conf file to control things like the maxKBps thruput settings.


docs/databags.md
```diff
@@ -114,6 +114,7 @@ Docs Navigation
 [server.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Serverconf
 [indexes.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Indexesconf
 [user-prefs.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/User-prefsconf
+[limits.conf]: http://docs.splunk.com/Documentation/Splunk/latest/admin/Limitsconf
```
acharlieh: This (and the change you made in the attributes doc) are just [markdown link references](http://daringfireball.net/projects/markdown/syntax#link) they don't make sense being added on their own without somewhere referring to them earlier in the doc.

recipes/_configure_limits.rb
```diff
@@ -0,0 +1,11 @@
+# coding: UTF-8
+#
+# Cookbook Name:: cerner_splunk
+# Recipe:: _configure_limits
+#
+# Configures the limits.conf settings for the system
+
+splunk_template 'system/limits.conf' do
+  stanzas node['splunk']['config']['limits']
```
acharlieh: 1) We probably need to default this attribute so that it's not required to be set to run.
2) If empty, should we remove the limits file? 
3) Does this belong under config or just under splunk (with apps and monitors)

docs/databags.md
bbrownz: Removed this line note, definitely doesn't make sense here. Updated the attributes.md with a note about the limits attributes.

95bc704

recipes/_configure_limits.rb
bbrownz: Added a default to nil (it actually wasn't required to be run previously, although the limits file does always get created). 

95bc704

I could add a check to ensure the file isn't created (or is removed) when there isn't content to populate it. It actually seems a bit safer to me to always create the file (potentially to have it be empty when attributes are removed). The removal portion I think is the tricky part, normal templates don't really behave that way, although they do have a delete action.

I think being under config makes sense, I might not be following though with the comparison to apps and monitors. Portions of the limits configuration apply to many of the different components in splunk.

acharlieh:
Closing per discussion on issue #63

PR #65, opened 2015-06-22
Fixes #64 - Tweaks to packaged app installation

### LINKED ISSUE #64: Issues with installing apps from tarballs
Oddness on Windows and on the Cluster Master we need to look into.

acharlieh:
To get this in, here are the fixes for windows installation of apps (i.e. do not chown on windows), and the fix for cluster apps (i.e. actually respect the download attributes)

There's also a rubocop update as well.

### DIFF
--- a/libraries/tarball.rb
+++ b/libraries/tarball.rb
@@ -5,6 +5,7 @@
 #
 # Methods for working with data from a tarball.
 
+require 'rubygems'
 require 'rubygems/package'
 require 'zlib'
 require 'fileutils'
@@ -29,6 +30,9 @@ module CernerSplunk
       @reader = @data[1] =  Gem::Package::TarReader.new(Zlib::GzipReader.open @file_name)
       @prefix = options[:prefix]
       @options = options
+      # chown blows up on windows in unexpected ways
+      # so give us a marker to know if we can chown
+      @can_chown = !Gem.win_platform?
     end
 
     attr_reader :prefix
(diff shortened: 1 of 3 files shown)


david-crowder:
+1

BobbyRyterski:
+1

libraries/tarball.rb
```diff
@@ -29,6 +30,7 @@ def initialize(file_name, options = {})
       @reader = @data[1] =  Gem::Package::TarReader.new(Zlib::GzipReader.open @file_name)
       @prefix = options[:prefix]
       @options = options
+      @chown = !Gem.win_platform?
```
garrypolley: For people coming at this code later it'd be nice if this had some doc on why it's used, and what it is doing. 

I'd also prefer it be something like `@can_chown`. That way the if checks read more like a sentence and are less ambiguous.

garrypolley:
Any chance for a spec to test this out? I'd expect this to have failed on windows previously and now it would pass.

libraries/tarball.rb
acharlieh: Renamed variable, and added a comment.

acharlieh:
Spec nope. I'm leveraging platform stuff (as being in a library fairly removed from access to the node object), so that means I'd need a VM (test-kitchen) to test upon. Figuring out how to do such correctly in as part of CI in the open is a much bigger story than getting this quick fix in and a couple of folks waiting on this change. (Yes, the code sucks in terms of tests)

garrypolley:
:+1:  the no specs is kind of a bummer but I get it.

PR #66, opened 2015-06-25
Fixes #64 - More Packaged app installation tweaks

### LINKED ISSUE #64: Issues with installing apps from tarballs

acharlieh:
Fixing #64 again. It turns out that [TarReader#close doesn't actually do anything](https://github.com/ruby/ruby/blame/v2_2_2/lib/rubygems/package/tar_reader.rb#L45-L49). 

As a result, we were not closing the file handle when we expected, so when we went to clean up the downloaded tarball on Windows we wound up with an exception since the file was still in use by us. (Linux this isn't as large of a problem since we keep the inode open, and the reference to the inode is removed from the file system without issue... but it could become a problem as we wind up with more open files than we expected)

### DIFF
--- a/libraries/tarball.rb
+++ b/libraries/tarball.rb
@@ -27,7 +27,9 @@ module CernerSplunk
       ObjectSpace.define_finalizer(self, @clean_proc)
       # Setup the reader, and read the prefix option if present.
       @file_name = @data[0] = file_name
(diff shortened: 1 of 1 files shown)


david-crowder:
+1

garrypolley:
:+1:

PR #74, opened 2015-07-17
Fixes #72 - Add Sensitive flag to License resource

### LINKED ISSUE #72: Missing sensitive flag on license files
It seems that the [license resources](https://github.com/cerner/cerner_splunk/blob/1.10.3/recipes/license_server.rb#L46-L52) is missing the sensitive flag. We should add it.

acharlieh:
Fixes #72 Apparently we missed adding the sensitive flag to the license resource. This will ensure that as new licenses are added we don't log out the contents.


BobbyRyterski:
:+1:

PR #76, opened 2015-08-06
Clean up Vagrant Artifacts

acharlieh:
Leading into #60, I just wanted to start with some cleanup of the Vagrant artifacts and documentation. 

Highlights: 
* Removal of the Chef 10 test box and related artifacts
* Removal of detection of Cerner network (and attempts to pull internal resources)
* Ensuring ChefDK is always the latest version (on a new provisioning of the chef image)

As this is all test code, I also propose no cookbook version increment upon merge. (please +/-1 this as well as the PR contents)

PR #79, opened 2015-08-17
Further Vagrantfile Fixes

acharlieh:
After #76 while working on the upgrade I realized a few things. One that Opentable Windows VM is missing a rather important line (Requested update from OpenTable here: https://twitter.com/acharlieh/status/629690576024137728 )

While I was at it, I realized that the default case shouldn't be to rebuild the apps all the time, so I added an ENV switch for it.

As it's Vagrant changes only, no version increment warranted.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -107,7 +107,7 @@ Vagrant.configure('2') do |config|
       cfg.vm.provision :shell, inline: 'cd /vagrant/vagrant_repo; mv .nodes.bak nodes', privileged: false
     end
 
-    cfg.vm.provision :shell, inline: <<-'SCRIPT'.gsub(/^\s+/, ''), privileged: false
+    app_gen = <<-'SCRIPT'.gsub(/^\s+/, '')
       mkdir -p "$HOME/app_service"
       rm -rf "$HOME/app_service/*"
       cd /vagrant/vagrant_repo/apps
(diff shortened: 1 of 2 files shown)


BobbyRyterski:
+1

Vagrantfile
```diff
@@ -123,10 +123,14 @@ Vagrant.configure('2') do |config|
         fi
       done
       cd "$HOME"
-      nohup /opt/chefdk/embedded/bin/ruby -run -e httpd "$HOME/app_service" -p5000 2>&1 > /dev/null &
+      netstat -nl | grep :5000 > /dev/null || nohup /opt/chefdk/embedded/bin/ruby -run -e httpd "$HOME/app_service" -p5000 2>&1 > /dev/null &
```
garrypolley: What exactly does doing the `netstat -nl | grep :5000 /dev/null ||` add here?

acharlieh: It looks to see if port 5000 is listening already. If it isn't then we attempt to start the ruby process that will listen on port 5000.

Basically this means that when we run `vagrant provision`, we don't try to start this ruby process over and over and over again.

garrypolley: I just learned what `||` does.  :fireworks:  Hadn't used it in bash before.

garrypolley:
:+1:

ibsharma:
+1

Vagrantfile
```diff
@@ -123,10 +123,14 @@ Vagrant.configure('2') do |config|
...
       sleep 10
     SCRIPT
 
+    app_gen.sub!('tar', '[ -f "$HOME/app_service/$D.tgz" ] || tar') unless ENV['REGEN_APPS']
```
nimeshsubramanian: should we add this env variable to our doc here:https://github.com/cerner/cerner_splunk/blob/stable/docs/vagrant.md ?

Vagrantfile
BobbyRyterski: Do we want to use `grep -q` instead of piping to `/dev/null`?

Vagrantfile
acharlieh: Added. We never documented the function of the chef VM being app hosting at all before either so added some verbiage to that effect.

Vagrantfile
acharlieh: True, while on the data set of all listening ports should be small enough that it shouldn't be an issue, it's a better practice.

nimeshsubramanian:
+1

PR #81, opened 2015-08-23
Fixes #80 Ensure .ui_login file is always created

### LINKED ISSUE #80: First time login and update password dialog showing up
It looks like this is broken
https://github.com/cerner/cerner_splunk/blob/1.10.3/recipes/_install.rb#L93-L96

Probably since 1.7.0 since it subscribes to the old name of the package resource
https://github.com/cerner/cerner_splunk/commit/4f68c7b98a8ab0732f6dd093f07bcb4847092a9e

As it's just a file that we touch, we should just touch the file, every time.

acharlieh:
Fixes #80. Instead of subscribing to a change, we're just going to ensure this file exists by touching it, but not if it already exists.

We probably should adopt this pattern with other resources too eventually.

Proposed: fix version increment as this is a bug.

### DIFF
--- a/recipes/_install.rb
+++ b/recipes/_install.rb
@@ -91,8 +91,8 @@ end
 
 # This gets rid of the change password prompt on first login
 file "#{node['splunk']['home']}/etc/.ui_login" do
-  action :nothing
-  subscribes :touch, "package[#{node['splunk']['package']['name']}]", :immediately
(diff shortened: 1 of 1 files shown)


aaronblythe:
:+1:

david-crowder:
+1

PR #82, opened 2015-08-25
Fixes #18 - Use a marker file for restart notifications

### LINKED ISSUE #18: Failure when using cerner_splunk_forwarder_monitors inside a separate run_context

acharlieh:
Fixes #18

Essentially we are implementing the design developed with https://github.com/cerner/cerner_splunk/pull/22#issuecomment-124629692 

But we create a marker file, and setup a delayed restart at the beginning. Every time we make a change, we touch the marker file. If we successfully start or restart, we delete the marker file.

Proposing a Minor version increment as while it fixes the expected state, it feels broader than just a simple fix.

### DIFF
--- a/libraries/recipe.rb
+++ b/libraries/recipe.rb
@@ -20,6 +20,10 @@ module CernerSplunk
     node.default['splunk']['node_type'] = symbol.to_sym
   end
 
+  def self.restart_marker_file
+    "#{Chef::Config[:file_cache_path]}/.restart_splunk"
+  end
+
   # Returns the key identifing the current cluster
   def self.my_cluster_key(node)
     node['splunk']['config']['clusters'].first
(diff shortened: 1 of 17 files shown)


acharlieh:
Ping: @tboettch and @mattnelson I know you were both interested in this functionality.

david-crowder:
+1

BobbyRyterski:
+1

nimeshsubramanian:
+1

PR #83, opened 2015-08-25
More Vagrant Tweaks - Resizing of vg00

acharlieh:
Doesn't actually do anything anymore, so we should remove it. 

Prior to #76, the Virtualbox images that we were using had a too small opt partition on vg00, which caused [Splunk's settings on free disk space](http://docs.splunk.com/Documentation/Splunk/6.2.5/Indexer/Setlimitsondiskusage#Set_minimum_free_disk_space) to kick off so we needed to resize it. The images we're using since that pull request, don't have a vg00, therefore this command is doing nothing but take a bit of time and spit out an error message which is ignored as we happily scoot along. 

Proposed no version increment as this is a Vagrantfile change only.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -34,11 +34,6 @@ def default_omnibus(config)
   config.omnibus.chef_version = :latest
 end
 
-def disk(config)
-  default_omnibus config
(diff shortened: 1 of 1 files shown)


ibsharma:
+1

nimeshsubramanian:
+1

Issue #52, opened 2015-03-15
Update to Splunk 6.2.5


acharlieh:
We need to look at Splunk 6.2.2 and update things as appropriate.

Of note, we need to look at how launching occurs as the limited user now, as well as how the new splunkweb process works. Double check directory permissions on upgrades as well.

acharlieh:
Interesting, perhaps change how we restart cluster peers. http://docs.splunk.com/Documentation/Splunk/6.2.5/Indexer/Restartthecluster

acharlieh:
Initial merge is in with fffed0132afa3ace4d83884d39798589689dae4b we'll leave this open in case there are any issues found during testing.

acharlieh:
We successfully updated one cluster, so going to close this for the current release

PR #84, opened 2015-08-31
Preliminary 6.2 

acharlieh:
Related to #52 We're changing things to make 6.2 install! Woot!

Proposed minor version bump.

### DIFF
--- a/recipes/_configure_outputs.rb
+++ b/recipes/_configure_outputs.rb
@@ -10,8 +10,8 @@ output_stanzas = {}
 if [:search_head, :forwarder, :cluster_master].include? node['splunk']['node_type']
   output_stanzas['tcpout'] = {
     'forwardedindex.0.whitelist' => '.*',
-    'forwardedindex.1.blacklist' => '_.*',
-    'forwardedindex.2.whitelist' => '(_audit|_internal)'
+    'forwardedindex.1.blacklist' => '_thefishbucket',
+    'forwardedindex.2.whitelist' => ''
   }
 
   # If we're part of a cluster, we only want to send events to our cluster.
--- a/recipes/_install.rb
+++ b/recipes/_install.rb
@@ -73,6 +73,13 @@ directory node['splunk']['external_config_directory'] do
   mode '0700'
 end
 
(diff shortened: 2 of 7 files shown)


BobbyRyterski:
+1

david-crowder:
+1

recipes/_configure_outputs.rb
```diff
@@ -10,8 +10,8 @@
 if [:search_head, :forwarder, :cluster_master].include? node['splunk']['node_type']
   output_stanzas['tcpout'] = {
     'forwardedindex.0.whitelist' => '.*',
-    'forwardedindex.1.blacklist' => '_.*',
-    'forwardedindex.2.whitelist' => '(_audit|_internal)'
```
aaronblythe: Why this change?  Can you provide links to the docs so I can learn more?

recipes/_install.rb
```diff
@@ -73,6 +73,13 @@ def nsp
   mode '0700'
 end
 
+# SPL-89640 On upgrades, the permissions of this directory is set incorrectly.
```
aaronblythe: As too permissive or not permissive enough? (could be a slightly better comment if it answers this question).

aaronblythe:
+1

recipes/_configure_outputs.rb
acharlieh: Added [in Splunk 6.1](http://docs.splunk.com/Documentation/Splunk/6.1/ReleaseNotes/MeetSplunk#Platform_instrumentation_framework) was the [Splunk Enterprise platform Instrumentation framework](http://docs.splunk.com/Documentation/Splunk/6.2.5/Troubleshooting/Abouttheplatforminstrumentationframework). This system by default on servers logs data to a new Splunk index called _introspection. As we're clustering indexers, we want to forward this data from search heads, and forwarders on which we enable this.

The way this is implemented as well (forward all indexes except for _thefishbucket, which isn't really an index but more of a scratch pad), additionally paves the way for the possibility of team separated  _* indexes, that way we could even start dividing up _internal data from forwarders for each team, and possibly lock down access to _internal itself.

nimeshsubramanian:
+1

PR #86, opened 2015-09-02
Check for empty hashes as well

### LINKED ISSUE #85: cerner_splunk_forwarder_monitors fails when passed empty monitor list
I get the error below when passing an empty list of monitors to ```cerner_splunk_forwarder_monitors```. Initial debugging suggests that an empty hash is passed to Chef's ```file.contents```, which is not permitted. The empty hash seems to be generated by ```CernerSplunk::LWRP.convert_monitors``` because there are no monitors to fold over.

My desired behavior is that any existing monitors be removed in this case, as if delete had been specified.

I have a proposed fix that I will submit shortly.

tboettch:
Fixes #85 

This seemed to be in agreement with the treatment of the empty string in the same method. Other solutions are possible.

### DIFF
--- a/libraries/splunk_app.rb
+++ b/libraries/splunk_app.rb
@@ -285,7 +285,7 @@ class Chef
           file.variables(stanzas: contents)
         else
           file = Chef::Resource::File.new(path, run_context)
-          file.content(contents)
+          file.content(contents) unless contents.empty?
         end
         file.path(path)
         file.owner(node['splunk']['user'])
--- a/libraries/splunk_app.rb
+++ b/libraries/splunk_app.rb
@@ -277,8 +277,8 @@ class Chef
       # function for dropping either a splunk template generated from a hash
       # or a simple file if the contents are a string. If the content of the file
       # is empty, then the file will be removed
-      def manage_file(path, contents)
-        if contents.class == Hash && contents.empty? == false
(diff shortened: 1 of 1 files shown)


tboettch:
The static analyzer seems to be complaining about something unrelated?

david-crowder:
"Perceived complexity for manage_file is too high" it looks like that extra conditional statement pushed rubocop over the edge for perceived complexity on that method. Possible solutions are disabling/increasing the cop (either on just that method or as a whole) or refactoring. I don't know which is the correct path in this instance, just thought I would give more info on the cause :)

tboettch:
I must have been misreading it because I thought it had identified another method altogether.

acharlieh:
Made a change to fix the travis run (ignore complexity for that method), and made another slight change regarding checking for child classes of Hash (like Mash!) ... I can't figure out a way to test this in it's current form (run_action adds pain... we probably need a refactor project at some point). But @tboettch if you could take a peek at acharlieh/cerner_splunk@d7672f30d82e172949f1c6a2414d7fdb43d981b1 and if you agree with it, pull that commit onto your stable branch that you're making this PR from, that'd be excellent.

tboettch:
I've merged Charlie's fixes in.

acharlieh:
:+1: Proposing fix version bump at merge.

david-crowder:
+1

BobbyRyterski:
+1

PR #89, opened 2015-09-10
Add recipe for deploying splunk as a heavy forwarder

### LINKED ISSUE #12: Recipe for configuring a Heavy Forwarder
AS A: Splunk Administrator
I WANT: The ability to setup Heavy Forwarders
SO THAT: I can offload parsing load from my Indexers, and potentially delegate control.

---

I think that #6 would help with the maintainability of this. #11 is likely a prerequisite for this.

jpeg:
~~Fixes #12~~
EDIT: I misunderstood the intent of #12, it is a different use case for a heavy forwarder.

Adds a new `cerner_splunk::heavy_forwarder` recipe to deploy a splunk heavy forwarder.

Tested using the vagrant deployments in this repo as well as by using the recipe to deploy in my own test setup and verifying that logs were forwarded to a local enterprise splunk instance.


docs/vagrant.md
```diff
@@ -11,15 +11,15 @@ Running with Vagrant
...
   1. Download the needed splunk packages locally, in a directory structure mirroring that of download.splunk.com
     * You can find the suffixes of files at the [Splunk download page](http://splunk.com/download)
-    * Copy a download link, it will look like `http://www.splunk.com/page/download_track?file=6.0.1/splunk/linux/splunk-6.0.1-189883-linux-2.6-x86_64.rpm&platform=...`
+    * Copy a download link, it will look like `http://www.splunk.com/page/download_track?file=6.2.5/splunk/linux/splunk-6.2.5-272645-linux-2.6-x86_64.rpm&platform=...`
     * The `file` query parameter will be the suffix you add onto `http://download.splunk.com/releases/`
-      * In this example the full url is: `http://download.splunk.com/releases/6.0.1/splunk/linux/splunk-6.0.1-189883-linux-2.6-x86_64.rpm`
-      * You will download the file locally to `<Mirror root>/releases/6.0.1/splunk/linux/splunk-6.0.1-189883-linux-2.6-x86_64.rpm
+      * In this example the full url is: `http://download.splunk.com/releases/6.2.5/splunk/linux/splunk-6.2.5-272645-linux-2.6-x86_64.rpm`
+      * You will download the file locally to `<Mirror root>/releases/6.0.1/splunk/linux/splunk-6.2.5-272645-linux-2.6-x86_64.rpm
```
acharlieh: Missed a version change, but also these changes have nothing to do with your proposed recipe addition. These should be a separate docs only PR.

acharlieh:
Your use case is you want to install and configure a full Splunk instance as a forwarder (instead of the Universal Forwarder)? I'll admit part of my thoughts on #12 was a Heavy Forwarder as part of offloading parsing load from indexers, and minimizing need for broadly open ports between networks as such cases arise. Thus they listen to ports, would have their own cluster data bag, and UFs could be configured to point to them instead of indexers and we could install apps to manage parsing before data is sent upstream to another cluster.

Other things going through my head include how do we figure out if someone is attempting to upgrade their node from a UF to HWF and do the right thing? (or how could people go the other way), and what is the right thing to do when we detect this case?

jpeg:
Yes, that is the use case I intended. As the UF doesn't include a bundled version of python a heavy forwarder would be the proper way to go for using python apps to do custom data collection. I suppose that what you intend with #12 and this probably should be handled by two separate recipes.

As for upgrading from one type to the other, do you think it would be sufficient to check whether `/opt/splunk` already exists when installing a UF or `/opt/splunkforwarder` for a HWF and then cleanup the existing forwarder if so?

docs/vagrant.md
jpeg: Removed the doc change from this PR: https://github.com/cerner/cerner_splunk/commit/56461c676a13ba1eefcca2cec4c3bc245ca8f2ea

acharlieh: Ideally, just squash it out of this branch, put it onto a different branch (instead of having a new commit. Adding to be removed a commit later on the same feature branch is just noise :)

jpeg: Done, created a separate PR for the doc change branch: https://github.com/cerner/cerner_splunk/pull/90

jpeg:
I've added cleanup of the existing forwarder when switching between a UF and HWF: https://github.com/cerner/cerner_splunk/commit/0d8aa910edd64565faae6f8bb0bfc7e88d7ecf74

However currently when migrating from UF -> HWF past log entries are being re-forwarded to the splunk server resulting in duplicates. I think that this is because the migration mechanism was intended for LWF -> UF and so that is the only route properly supported. I also tried manually copying the fishbucket at various points in the install process to the new instance but that method had the same result.

acharlieh:
Closing this PR in favor of #113.

PR #98, opened 2015-12-02
Fixes #96. Add ulimit command to init.d script to ensure ulimits are applied

### LINKED ISSUE #96: Ensure ulimits are set appropriately post 6.2
It seems that after 6.2 that the ulimit value is no longer being respected, that is because instead of the generated init script becoming the limited user with su, and then launching the process (thus applying limits.conf appropriately), the method now is where the process launches as root and then the process nerfs itself to the limited user (thus maintaining the ulimits of the launching shell).

We need to figure out the best way around this issue and ensure our cookbook sets appropriate ulimits again.

jdmurphy:
Updated _start recipe to edit the splunk init.d script with a ulimit command if needed.

### DIFF
--- a/recipes/_restart_marker.rb
+++ b/recipes/_restart_marker.rb
@@ -1,0 +1,13 @@
+# coding: UTF-8
+#
+# Cookbook Name:: cerner_splunk
+# Recipe:: _restart_marker
+#
+# Creates the restart file marker. This is used to ensure that the service is only restarted once and that it
+# only restarts when needed
+
+file 'splunk-marker' do
+  action :nothing
+  backup false
+  path CernerSplunk.restart_marker_file
+end
--- a/recipes/_start.rb
+++ b/recipes/_start.rb
@@ -5,9 +5,28 @@
 #
 # Ensures the splunk instance is running, and performs post-start tasks.
 
+ulimit_command = "ulimit -n #{node['splunk']['limits']['open_files']}"
+init_file_path = '/etc/init.d/splunk'
+restart_flag = !(File.exist?(init_file_path) && File.readlines(init_file_path).grep(/#{ulimit_command}/).any?)
+
 # We want to always ensure that the boot-start script is in place on non-windows platforms
 execute "#{node['splunk']['cmd']} enable boot-start -user #{node['splunk']['user']}" do
   not_if { platform_family?('windows') }
+end
+
+ruby_block 'insert ulimit' do
+  block do
+    file = Chef::Util::FileEdit.new(init_file_path)
+    file.insert_line_after_match(/^RETVAL=\d$/, ulimit_command)
+    file.write_file
+  end
+  not_if { platform_family?('windows') }
+end
+
+ruby_block 'restart-splunk-for-ulimit' do
+  block { true }
+  notifies :touch, 'file[splunk-marker]', :immediately
+  only_if { !platform_family?('windows') && restart_flag }
 end
 
 # We then start splunk. In the future, the other resource should be here instead of this clumsy notification
--- a/spec/unit/recipes/_restart_marker_spec.rb
+++ b/spec/unit/recipes/_restart_marker_spec.rb
@@ -1,0 +1,18 @@
+# coding: UTF-8
+
+require_relative '../spec_helper'
+
+describe 'cerner_splunk::_restart_marker' do
+  subject do
+    runner = ChefSpec::SoloRunner.new
+    runner.converge(described_recipe)
+  end
+
+  before do
+    allow(CernerSplunk).to receive(:restart_marker_file).and_return('/foo/bar')
+  end
+
+  it 'does nothing with the splunk marker file' do
+    expect(subject.file('splunk-marker')).to do_nothing
+  end
+end
--- a/spec/unit/recipes/_start_spec.rb
+++ b/spec/unit/recipes/_start_spec.rb
@@ -1,0 +1,145 @@
+# coding: UTF-8
+
+require_relative '../spec_helper'
+
+describe 'cerner_splunk::_start' do
+  subject do
+    runner = ChefSpec::SoloRunner.new(platform: platform, version: platform_version) do |node|
+      node.set['splunk']['cmd'] = 'splunk'
+      node.set['splunk']['user'] = 'splunk'
+      node.set['splunk']['config']['clusters'] = ['cerner_splunk/cluster']
+    end
+    # Have to include forwarder recipe so that _start recipe can send notifications to services
+    runner.converge('cerner_splunk::forwarder', described_recipe)
+  end
+
+  let(:cluster_config) do
+    {
+      cluster: {
+        receivers: ['33.33.33.20'],
+        license_uri: nil,
+        receiver_settings: {
+          splunktcp: {
+            port: '9997'
+          }
+        },
+        indexes: 'cerner_splunk/indexes'
+      }
+    }
+  end
+
+  let(:platform) { nil }
+  let(:platform_version) { nil }
+
+  let(:lines) { [] }
+  let(:exists) { nil }
+
+  let(:windows) { nil }
+
+  before do
+    allow(Chef::DataBagItem).to receive(:load).with('cerner_splunk', 'cluster').and_return(cluster_config)
+    allow(Chef::Recipe).to receive(:platform_family?).with('windows').and_return(windows)
+
+    allow(File).to receive(:exist?).and_call_original
+    allow(File).to receive(:exist?).with('/etc/init.d/splunk').and_return(exists)
+    allow(File).to receive(:readlines).and_call_original
+    allow(File).to receive(:readlines).with('/etc/init.d/splunk').and_return(lines)
+
+    # Stub alt separator for windows in Ruby 1.9.3
+    stub_const('::File::ALT_SEPARATOR', '|')
+  end
+
+  context 'when platform is windows' do
+    let(:platform) { 'windows' }
+    let(:platform_version) { '2012R2' }
+    let(:windows) { true }
+
+    before do
+      ENV['PROGRAMW6432'] = 'test'
+    end
+
+    it 'does not execute boot-start script' do
+      expect(subject).to_not run_execute('splunk enable boot-start -user splunk')
+    end
+
+    it 'does not insert ulimit into init.d script' do
+      expect(subject).to_not run_ruby_block('insert ulimit')
+    end
+
+    it 'does not run restart-splunk-for-ulimit ruby block' do
+      expect(subject).not_to run_ruby_block('restart-splunk-for-ulimit')
+    end
+  end
+
+  context 'when platform is not windows' do
+    let(:platform) { 'centos' }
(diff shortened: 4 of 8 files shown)


jdmurphy:
Travis failed for an error related to chef-cookbooks/chef-vault/issues/41

recipes/_restart_marker.rb
```diff
@@ -0,0 +1,13 @@
+# coding: UTF-8
+#
+# Cookbook Name:: cerner_splunk
+# Recipe:: _restart_marker
+#
+# Creates the restart file marker. This is used to ensure that the service is only restarted once and that it
+# only restarts when needed
+
+file 'splunk-marker' do
```
gravesb: was there a particular reason this resource was pulled out into a separate recipe? Was it just so we could unit test it easier?

recipes/_start.rb
```diff
@@ -10,6 +10,16 @@
   not_if { platform_family?('windows') }
 end
 
+ruby_block 'insert ulimit' do
+  block do
+    file = Chef::Util::FileEdit.new('/etc/init.d/splunk')
+    file.insert_line_after_match(/^RETVAL=\d$/, "ulimit -n #{node['splunk']['limits']['open_files']}")
+    file.write_file
+  end
+  not_if { platform_family?('windows') || File.readlines('/etc/init.d/splunk').grep(/ulimit/).any? }
```
gravesb: What if we want to adjust the value of the ulimit?  If the current value is 8196 and I update my config to a different value, this logic won't every modify the ulimit in the init.d file again.

recipes/_restart_marker.rb
jdmurphy: That was definitely part of it. It was already a bit of a shared resource.

recipes/_start.rb
jdmurphy: Actually I could probably get rid of the greping for ulimit because, and Charlie can correct me if I am wrong, I believe the init.d script gets overwritten on each deploy so it would never have the ulimit added in the script.

recipes/_start.rb
```diff
@@ -10,6 +10,16 @@
   not_if { platform_family?('windows') }
 end
 
+ruby_block 'insert ulimit' do
+  block do
+    file = Chef::Util::FileEdit.new('/etc/init.d/splunk')
+    file.insert_line_after_match(/^RETVAL=\d$/, "ulimit -n #{node['splunk']['limits']['open_files']}")
```
gravesb: What are the odds that splunk would every modify the first line of the init.d file to something other than `RETVAL=\d`?  I would be more comfortable with this if we had a serverspec to cover it in an integration test.  Otherwise we may be better off doing something like [this](http://stackoverflow.com/questions/8623231/prepend-a-single-line-to-file-with-ruby) to simply prepend that line to the file.

jdmurphy: The problem is we can't simply prepend it since the line has to come after the magic bash comment that is the first line of the script.

recipes/_start.rb
acharlieh: So you're correct in that the boot-start command re-writes the file every time, so this would edit the file every time.

But that turns line 20 into a problem, as we'd only want to mark the service for a restart if the ulimit has changed. (we'd wind up restarting every time here)... Possibly have to look at the file and set the marker beforehand.

jdmurphy: Might it be better just not to recreate the init script every time?

jdmurphy:
Just a note I will squash this down more once the review part is wrapped up.

spec/unit/recipes/_start_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+        end
+
+        context 'when ulimit command is different than the one to be added' do
+          let(:lines) { ['line 1', 'ulimit -n 1234'] }
+
+          it 'inserts ulimit into init.d script' do
+            expect(subject).to run_ruby_block('insert ulimit')
+          end
+
+          it 'does not notify the touch restart-marker resource' do
```
gravesb: it 'notifies the touch ...'

recipes/_start.rb
```diff
@@ -5,11 +5,38 @@
 #
 # Ensures the splunk instance is running, and performs post-start tasks.
 
+def restart_needed?(ulimit_command)
```
gravesb: minor, thoughts on just doing this inline rather than putting it in a function?
````
ulimit_command = "..."
init_file_path = '...'
restart_needed = !File.exist(init_file_path) || File.readlines(...).none?
````
Then the file_path could be reused down on line 27.

**Comment by nabrmatt** [comment]
+1 after squashing appropriate commits together.

File: recipes/_start.rb [hunk 6]
```diff
@@ -5,11 +5,30 @@
...
+ruby_block 'insert ulimit' do
+  block do
+    file = Chef::Util::FileEdit.new(init_file_path)
+    file.insert_line_after_match(/^RETVAL=\d$/, ulimit_command)
+    file.write_file
+  end
+  not_if { platform_family?('windows') }
+end
+
+ruby_block 'restart-splunk' do
```
acharlieh: I think this resource needs to be before the execute resource, as guards are evaluated during the execution phase so this will evaluate after we've blown away the file, and made the ulimit edit (and thus never notify).

Also we should name the resource check-for-ulimit-change or something restart-splunk is very generic.

recipes/_start.rb
```diff
@@ -5,11 +5,30 @@
...
+  block { true }
+  notifies :touch, 'file[splunk-marker]', :immediately
+  only_if { restart_flag }
```
acharlieh: This guard needs adjustment as well, because windows

We want to notify for restart:
not_if (windows OR ( init file exists AND correct ulimit setting is present prior to editing))

spec/unit/recipes/_restart_marker_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+require_relative '../spec_helper'
+
+describe 'cerner_splunk::_restart_marker' do
+  subject do
+    runner = ChefSpec::SoloRunner.new
+    runner.converge(described_recipe)
+  end
+
+  before do
+    allow(CernerSplunk).to receive(:restart_marker_file).and_return('/opt/splunk')
```
acharlieh: I see what you're doing here and it doesn't matter too much but '/opt/splunk' is more than likely the install directory, the marker file is more likely actually '/var/chef/cache/.restart_splunk' 

It could be style, but I would want to either have mock values that are either likely real values, or values (to someone who doesn't know much about Splunk installation) that are very obviously wrong as I don't want tests to add confusion.

spec/unit/recipes/_start_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+
+  context 'when platform is windows' do
+    let(:platform) { 'windows' }
+    let(:platform_version) { '2012R2' }
+    let(:windows) { true }
+
+    before do
+      ENV['PROGRAMW6432'] = 'test'
+    end
+
```
acharlieh: Probably add a case that this restart block does not notify the marker.

recipes/_start.rb
jdmurphy: I am not sure I see the issue with the guard. Since the variable we are using in the guard is just plain ruby code, it should be evaluated at compile time so we should be checking if a change actually occurred between the past state and what will be applied in the insertion. 

As for renaming it I am fine with that, but it is touching the restart marker so restart-splunk seems to make more sense than check-for-ulimit-change since that is really just what the guard is doing. I am open to suggestions for the name of the resource.

gravesb: I think the placement is fine. Guards are evaluated during the execution phase, but in this case the value of the restart_flag is populated during the compile phase. Although I don't think it would hurt anything to move it above the boot-start execution resource.

acharlieh: Oh yep, you're right about that aspect of the guard... I was inlining the execution of the ruby code in my head

How about restart-splunk-for-ulimit-update ?

recipes/_start.rb
gravesb: Wouldn't the restart flag just always be false for windows (because the file will never exist)?

acharlieh: It (was) wrapped in a not, which makes it true when the file doesn't exist. (additionally when the file doesn't exist on linux the restart flag from this aspect should be true because ulimit isn't set out of the box. )

gravesb: ah, yep. I fail at boolean logic today.

acharlieh: It's ok, I've apparently been failing at execution order so cheers! :beers:

acharlieh:
I'm +1 for the final state... but one more pass at the history before merging. 

I'd propose these commits in order:
# 1e0017a923c483335ce66bb2a911769c6a739023
# The changes moving restart marker from 5b4bffbb5fb8b001f766bbd31f6b776ad84cbacc + the spec for it from 4f03ade14e037e34f5498ec87021c3c86cf1d17c
# 40acba005418842d6dfb5f6970508c376640b340 + The change to _start from  5b4bffbb5fb8b001f766bbd31f6b776ad84cbacc + the spec for _start from 4f03ade14e037e34f5498ec87021c3c86cf1d17c

This way history has you making three logical steps that build on eachother in this branch, no introducing and editing the same changes within a single merge :) (I know I might not be best at this all the time, but it's the goal).

jdmurphy:
@acharlieh is that better for history?

acharlieh:
@jdmurphy excellent

I'm good with merging them as is if you are. The next level of git-fu is then multi-line commit messages... where your first line is short enough to fit on a line and gets the gist across, but the other lines of the message are there with additional details. @david-crowder did a good example in 4e5ba715f6fd975d15f711d3566a8b5e02529074 and I'll freely admit that sometimes I do this well but sometimes my first line runneth over too (e.g. 1a53e51338f873f3fc6f80fbc4446afc1d968cca ) ... So definitely feel free to tell me to take a hike on this if you want.

The message for 069f3633836858dffbfbc50b45587fc0565da04f could be:

    Locked chef-vault cookbook to v1.3.0.

    Ruby 2.x specific syntax was added later and we need to support Chef 11 for now.
    Also pulled gem into Gemfile for easier testing.

b40e2d7f9f4647b12b1183a2a7d01d4ab0495bfe could be:

    Broke out restart marker into it's own recipe.

    This will enable easier testing of things needing to notify the restart marker.
    Also Look! Actual Specs! /me kicks @acharlieh :D :see_no_evil: :hear_no_evil: :speak_no_evil:

4ae7e143132da701db74d11cfa57a2a8f68bf796 could be:

    Set ulimit with service startup.

    With 6.2, the generated init script no longer uses `su` and thus doesn't pick up limits.d. 
    So we ensure that ulimit is set appropriately with the startup script, and if the value changes we ensure we restart.

gravesb:
+1

jdmurphy:
@acharlieh, no, it's good. I should be less lazy with my commit messages. I updated the wording.

acharlieh:
Referencing #96 in a comment. For whatever reason Github doesn't make a linkback based on a PR title but does when it appears in the body of a comment.

PR #100, opened 2016-02-18
Allow convenient specification of volumes for index storage locations

gravesb:
This allows a convenient way to specify the base volume for a given index.  Otherwise you would have to specify the coldPath and homePath for each index individually within the indexes data bag.  I originally considered putting the volume within the config/\<index_name\> location but then that hash would no longer match the raw indexes.conf stanza anymore.  So I opted to put it in the flags, although I know those were originally meant to be boolean only.  I'm open to suggestion on a better place to put the volume specification for an index in the data bag if you want to keep flags as only boolean attributes.

### DIFF
--- a/docs/databags.md
+++ b/docs/databags.md
@@ -37,7 +37,9 @@ Indexes Hash
 ------------
 An Indexes Hash is part of a plaintext data bag item that defines the set of indexes defined in a cluster. It is separate from the Cluster data bag primarily for size concerns.
 
-* `['config']` - These define the [indexes.conf] stanzas (in fairly raw form).
+* `['config']` - These define the [indexes.conf] stanzas (in fairly raw form). However there are a few special keys:
+    * `_volume` - The base volume for the coldPath, homePath and tstatsHomePath. Defaults to nil, so the index will be located in $SPLUNK_DB.
+    * `_directory_name` - The on-disk name of the directory to store the index. Defaults to the index name.
 * `['flags']` - These define boolean processing flags per index. All flags are default 'false' but can be set to true. Current flags include:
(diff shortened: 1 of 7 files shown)


acharlieh:
Yeah, the idea of flags was to keep config part close to indexes.conf, but in hindsight, I wonder if it would have been better to just prefix all "flag" fields as "_" so that way their configurations would be inline with everything else (and not have two parts of a data bag to look at... ). If we want to start this with the volume setting, I'd be down with that :) )

acharlieh:
Two other things that we may want to consider for configurability of indexes.conf... 
1) a flag for specifying the on-disk name for an index... this might be useful for if people may want to rename indexes, but also for being able to easily use the volume specification. Additionally for built in splunk indexes (e.g. _internal, main, _audit) who for legacy reasons use on disk names different from their index names (e.g. _internaldb, defaultdb, audit). (we could default this to the name of the index of course).
2) additionally managing the `tstatsHomePath` parameter... by default this is set for any new index to `volume:_splunk_summaries/$_index_name/datamodel_summary` which maps by default to $SPLUNK_DB, but if we're relocating indexes to alternate volumes, I think we'd also want to relocate datamodel summaries as well?

gravesb:
I have all the suggested changes updated on the PR, although it looks like ruby 1.9.3 doesn't like my custom chefspec matcher :(

gravesb:
Fixed the custom chefspec matcher to work with both Chef 11 and Chef 12 and rebased on stable to include the rubocop fixes. The builds are all passing now for this branch.

jdmurphy:
+1

alecxvs:
+1

PR #104, opened 2016-03-06
Make the source and issues url optional.

acharlieh:
These weren't added until Chef 12, so we need to have them be optional for as long as we support Chef 11.
They were introduced with: chef/chef@fe6d60f7a9456a97457f37a20729e366a714f428

Proposing no version change on merge (as it's nothing to do with the cookbook itself, just the metadata of it.


gravesb:
+1

gseats:
+1

PR #111, opened 2016-04-04
Fix chef-dk install command in Vagrantfile

### LINKED ISSUE #110: Chef Vagrant VM fails to start
Currently if you run `vagrant up chef`, it will fail with the following output:

```
==> chef: package chefdk is not installed
==> chef:  
==> chef:  
==> chef: %
==> chef:  
==> chef: T
==> chef: o
==> chef: t
==> chef: a
==> chef: l
==> chef:  
==> chef:  
==> chef:  
==> chef:  
==> chef: %
==> chef:  
==> chef: Received % Xferd  Average Speed   Time    Time     Time  Current
==> chef:                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
==> chef: 1
==> chef: 0
==> chef: 0
==> chef:  
==> chef: 16494  100 16494    0     0  69417      0 --:--:-- --:--:-- --:--:--  383k
==> chef: Downloading Chef  for el...
==> chef: downloading http://opscode.com/chef/metadata-chefdk?v=&prerelease=false&nightlies=false&p=el&pv=6&m=x86_64
==> chef:   to file /tmp/install.sh.3240/metadata.txt
==> chef: trying wget...
==> chef: sha1	619dfe23e23b0af622052827f564a59beded7973
==> chef: sha256	66c8b0e53c8512df62d7ef7f1dcd93fddd9ec23853d2ee22ce63f7173c12a7c3
==> chef: url	https://packages.chef.io/stable/el/6/chefdk-0.12.0-1.el6.x86_64.rpm
==> chef: version	0.12.0
==> chef: downloaded metadata file is corrupted or an uncaught error was encountered in downloading the file...
==> chef: Please file a bug report at http://tickets.opscode.com
==> chef: Project: Chef
==> chef: Component: Packages
==> chef: Label: Omnibus
==> chef: Version: 
==> chef:  
==> chef: Please detail your operating system type, version and any other relevant details
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

I looked at the [installer script](https://chef.sh/) we were using which had a lovely comment for the metadata file being corrupt message we got:

> this generally means one of the download methods downloaded a 404 or something like that and then reported a successful exit code,
>  and this should be fixed in the function that was doing the download.

Looking further into the script, it is still using old opscode URLs which when I attempt to curl them, first 301, and then try redirecting to chef.io which causes a 404.

jdmurphy:
This PR is to fix #110 where the chef-dk install fails when attempting to run vagrant. I based the new install command on the [chef-dk's readme installation instructions](https://github.com/chef/chef-dk#linux-and-osx)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -79,7 +79,7 @@ Vagrant.configure('2') do |config|
   config.vm.define :chef do |cfg|
     config.omnibus.chef_version = nil
 
-    cfg.vm.provision :shell, inline: 'rpm -q chefdk || curl -L https://chef.sh | bash -s -- -P chefdk'
+    cfg.vm.provision :shell, inline: 'rpm -q chefdk || curl -L https://omnitruck.chef.io/install.sh | bash -s -- -P chefdk'
 
     if ENV['KNIFE_ONLY']
       cfg.vm.provision :shell, inline: 'cd /vagrant/vagrant_repo; mv nodes .nodes.bak', privileged: false


gravesb:
+1

Vagrantfile
```diff
@@ -79,7 +79,7 @@ Vagrant.configure('2') do |config|
   config.vm.define :chef do |cfg|
     config.omnibus.chef_version = nil
 
-    cfg.vm.provision :shell, inline: 'rpm -q chefdk || curl -L https://chef.sh | bash -s -- -P chefdk'
+    cfg.vm.provision :shell, inline: 'curl https://omnitruck.chef.io/install.sh | bash -s -- -c current -P chefdk'
```
david-crowder: Is there a reason to drop the rpm call? This seems to try and install every time you provision now.

jdmurphy: Nope, just a mistake. Fixed ba093a6

acharlieh:
-1 We need to talk to Chef, Inc. before making this change because yet again this Chef being non-passive in their URL deprecation. 

They previously documented the chef.sh url [as a valid Omnitruck URL at least on their blog](https://www.chef.io/blog/2015/07/16/5-ways-to-deal-with-the-install-sh-curl-pipe-bash-problem/). I think I remember it being in the docs as well but of course that'll take digging to find it in the docs history, assuming it was in the new docs and wasn't just on the old wiki docs. 

Regarding the other changes... we need `-L` with curl, if Chef properly deprecated the URL, they would have left a redirect. 

Additionally I don't think we want `-c current` per their docs this tells omnitruck to use the prerelease channel. I'm pretty sure that we just want the latest stable (which is default).

acharlieh:
+1 Going to do green button merge (as this is only a change to vagrant, no cookbook code changed here).

PR #114, opened 2016-05-05
Relaxed chef-vault constraint

### LINKED ISSUE #112: Update/Remove chef-vault dependency constraint
The [issue](https://github.com/cerner/cerner_splunk/blob/stable/metadata.rb#L12-L14) with chef-vault requiring ruby 2 is now fixed in [1.3.3](https://github.com/chef-cookbooks/chef-vault/commit/e48d4b561dad36fbc2dfb4f66c3c3e6616b9e758)

bbaugher:
Fixes #112

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -9,9 +9,7 @@ version          '1.17.0'
 source_url       'https://github.com/cerner/cerner_splunk' if defined?(:source_url)
 issues_url       'https://github.com/cerner/cerner_splunk/issues' if defined?(:issues_url)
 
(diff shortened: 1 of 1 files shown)


acharlieh:
It looks like the Ruby 1.9.3 (Chef 11) build failed with this change... It is indeed only in test code as it's only executed if ChefSpec is defined, but I'm hesitant to introduce a change that prevents us from having at least our very basic tests pass. Or we need to discuss dropping support for Chef 11.

PR #115, opened 2016-05-12
Splunk 6.3.4

gravesb:
Update default Splunk version attributes to install Splunk 6.3.4.

Also tested upgrade path from previous version by provisioning vagrant before and after switching the version attributes.


nmattam:
+1

acharlieh:
+1

PR #116, opened 2016-05-16
splunk_app resource to support lookup specification

nmattam:
This PR is to enhance the splunk_app resource to support lookup specification. 

The use case is that we need to deploy a lookup file to the license server so that we can have a index-ownership lookup which can potentially help in notifying the index owners about any over indexing that might lead to license violation. This enhancement will help us source control the contents of the lookup file.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -106,6 +106,8 @@ Vagrant.configure('2') do |config|
     app_gen = <<-'SCRIPT'.gsub(/^\s+/, '')
       mkdir -p "$HOME/app_service"
       rm -rf "$HOME/app_service/*"
+      cd /vagrant/vagrant_repo/files
+      cp -R lookups "$HOME/app_service/"
       cd /vagrant/vagrant_repo/apps
       timestamp=`date -u +%Y%m%d%H%M%S`
       for D in *; do
--- a/docs/databags.md
+++ b/docs/databags.md
@@ -109,6 +109,7 @@ An apps hash is a contextual (see above) Hash, part of a plaintext data bag item
 * `[app]['download']['version']` - Expected [version number][app.conf] (if any) used to determine if a new app should be downloaded.
 * `[app]['files']` - Hash of files to manage under the "default" or "local" directory.
 * `[app]['files'][filename]` - Contents of a particular file to manage. It can take 3 values, a hash of stanzas -> key-value pairs (then written with the splunk template), a string (written as is), or nil / false (deleted). If the hash or string is empty, the file is also deleted.
+* `[app]['lookups']` - Hash of lookup files for the app. The key in the hash will be the name of file when it lands in the Splunk app and the value will be the url to lookup file. To delete an existing lookup file, set the value of the lookup file to `false` or `null` or to an empty string. The only supported file name extensions are .csv, .csv.gz and .kmz. Please see [splunk docs](http://docs.splunk.com/Documentation/Splunk/6.3.2/Knowledge/Addfieldsfromexternaldatasources) for more information about the supported file formats. `[app]['files']` hash can be used to specify any .conf setting that is required for the lookup.
 * `[app]['permissions']` - Hash of permissions to manage for the app.
 * `[app]['permissions'][object]` - Permissions to manage for a particular knowledge object or class of knowledge objects
 * `[app]['permissions'][object]['access']['read']` - array of roles or String '*' allowed to read the object
--- a/libraries/splunk_app.rb
+++ b/libraries/splunk_app.rb
@@ -58,6 +58,7 @@ class Chef
         @allowed_actions = [:create, :remove]
         @local = false
         @permissions = {}
+        @lookups = {}
         @files = {}
         @app = name
       end
@@ -73,6 +74,10 @@ class Chef
 
       def files(arg = nil)
         set_or_return(:files, arg, kind_of: Hash)
+      end
+
+      def lookups(arg = nil)
+        set_or_return(:lookups, arg, kind_of: Hash)
       end
 
       def permissions(arg = nil)
@@ -106,6 +111,10 @@ class Chef
 
       def files_dir
         local ? "#{root_dir}/local" : default_dir
+      end
+
+      def lookup_dir
+        "#{root_dir}/lookups"
       end
 
       def perms_file
@@ -143,6 +152,7 @@ class Chef
         download_and_install
         create_app_directories unless new_resource.updated_by_last_action?
         manage_metaconf unless new_resource.permissions.empty?
+        manage_lookups unless new_resource.lookups.empty?
         new_resource.files.each do |file_name, contents|
           *directories, file_name = file_name.split('/')
           file_path = new_resource.files_dir
@@ -202,10 +212,7 @@ class Chef
 
         filename = "#{Chef::Config[:file_cache_path]}/#{new_resource.app}.tgz"
 
-        download = Chef::Resource::RemoteFile.new(filename, run_context)
-        download.source(new_resource.url)
-        download.run_action(:create)
-        download.backup(false)
+        download = download_file filename, new_resource.url
 
         install_from_tar filename, expected_version, installed_version
       ensure
@@ -266,6 +273,18 @@ class Chef
         tarfile.close if tarfile
       end
 
+      def manage_lookups
+        lookups = new_resource.lookups
+        lookups.each do |file_name, url|
+          if url && !url.empty?
+            fail "Unsupported lookup file format for #{file_name} in the app #{new_resource.app}" unless file_name =~ /\.(?:csv\.gz|csv|kmz)$/i
+            download_file ::File.join(new_resource.lookup_dir, file_name), url
+          else
+            delete_file ::File.join(new_resource.lookup_dir, file_name)
+          end
+        end
+      end
+
       def manage_metaconf
         permissions = new_resource.permissions
         permissions.each do |stanza, hash|
@@ -276,6 +295,18 @@ class Chef
           end
         end
         manage_file(new_resource.perms_file, permissions)
+      end
+
+      def download_file(file_path, url)
+        download = Chef::Resource::RemoteFile.new(file_path, run_context)
+        download.source(url)
+        download.backup(false)
+        download.run_action(:create)
+      end
+
+      def delete_file(file_path)
+        download = Chef::Resource::File.new(file_path, run_context)
+        download.run_action(:delete)
       end
 
       # function for dropping either a splunk template generated from a hash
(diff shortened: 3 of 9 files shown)


acharlieh:
Looking at the design, out of curiosity, is there a reason you went with having the data in the data bag, as opposed to say, pointing to a URL of where to download the lookup? (Alternatively / additionally, allowing a lookup to be removed if the key exists with nil or false as the value?)

nmattam:
I thought of managing the lookups in the same way we manage other files/objects in the app using a databag. Assuming we have one chef repository which maintains all the different databags, this would make updating the databag with the index-owner lookup easier (which was the specific use case here) when we create a new index by avoiding the maintenance of a separate file outside the chef repository.

I see that there is no way to remove a lookup file once deployed in to the app other than manually deleting it. I will update the PR to support removing of lookup files.

nmattam:
Updated the PR based on the Charlie's last comment.
Changes made:
1. Specifying the lookup file as URL rather than having a databag with the lookup.
2. Option to delete a lookup.
3. Removed the databag that was supposed to install apps on the license server. Instead added the lookup specification to the awesome_splunk_app that will be installed on the standalone instance.

libraries/splunk_app.rb
```diff
@@ -278,6 +297,19 @@ def manage_metaconf
         manage_file(new_resource.perms_file, permissions)
       end
 
+      def download_file(file_path, url)
+        download = Chef::Resource::RemoteFile.new(file_path, run_context)
+        download.source(url)
+        download.run_action(:create)
+        download.backup(false)
```
acharlieh: I think this needs to be before the previous line to have any effect... run_action actually invokes the action to be performed, and then you setting an attribute afterwards.

See: https://www.chef.io/blog/2013/09/04/demystifying-common-idioms-in-chef-recipes/

Vagrantfile
```diff
@@ -104,8 +104,10 @@ Vagrant.configure('2') do |config|
     end
 
     app_gen = <<-'SCRIPT'.gsub(/^\s+/, '')
-      mkdir -p "$HOME/app_service"
+      mkdir -p "$HOME/app_service/lookups"
```
acharlieh: I don't think this change is actually doing anything because we do an `rm -rf `on the parent directory immediately afterwards.

docs/databags.md
```diff
@@ -109,6 +109,7 @@ An apps hash is a contextual (see above) Hash, part of a plaintext data bag item
 * `[app]['download']['version']` - Expected [version number][app.conf] (if any) used to determine if a new app should be downloaded.
 * `[app]['files']` - Hash of files to manage under the "default" or "local" directory.
 * `[app]['files'][filename]` - Contents of a particular file to manage. It can take 3 values, a hash of stanzas -> key-value pairs (then written with the splunk template), a string (written as is), or nil / false (deleted). If the hash or string is empty, the file is also deleted.
+* `[app]['lookups']` - Hash of lookup files for the app. The key in the hash will be the name of file when it lands on splunk and value will be the url to lookup file. To delete an existing lookup file, set the value of the lookup file to empty or to `"nil"`. Only supported file name formats are .csv, .gz and .kmz. If you are downloading a comma separated text file (.txt), the file name you specify as the key in the hash should have a .csv extension.
```
acharlieh: > Only supported file name formats are .csv, .gz and .kmz. If you are downloading a comma separated text file (.txt), the file name you specify as the key in the hash should have a .csv extension.

Where is this requirement from? How is it being enforced?

> To delete an existing lookup file, set the value of the lookup file to empty or to `"nil"`.

To the string nil ? and not just boolean false or null?

libraries/splunk_app.rb
```diff
@@ -266,6 +273,18 @@ def install_from_tar(filename, expected_version, installed_version)
         tarfile.close if tarfile
       end
 
+      def manage_lookups
+        lookups = new_resource.lookups
+        lookups.each do |file_name, url|
+          if url == 'nil' || url.empty?
```
acharlieh: We should not have magic strings.

libraries/splunk_app.rb
```diff
@@ -278,6 +297,19 @@ def manage_metaconf
...
 
+      def download_file(file_path, url)
+        download = Chef::Resource::RemoteFile.new(file_path, run_context)
+        download.source(url)
+        download.run_action(:create)
+        download.backup(false)
+      end
+
+      def delete_file(file_path)
+        return unless ::File.exist? file_path
```
acharlieh: Why is this necessary? is `:delete` not idempotent?

libraries/splunk_app.rb
nmattam: Fixed it here: 8a626089e661f787e149847348f693986cd49e45

docs/databags.md
nmattam: Saw this in the lookup upload view in the splunk web UI
>Select either a plaintext CSV file, a gzipped CSV file, or a KMZ file. 
The maximum file size that can be uploaded through the browser is 500MB.
Enter the name this lookup table file will have on the Splunk server. If you are uploading a gzipped CSV file, enter a filename ending in ".gz". If you are uploading a plaintext CSV file, we recommend a filename ending in ".csv". For a KMZ file, we recommend a filename ending in ".kmz".

Thought of giving the value always the same type. Agreed that it was not a good idea. Updated the option to be null if the lookup needs to be deleted.  8a626089e661f787e149847348f693986cd49e45

libraries/splunk_app.rb
nmattam: 8a626089e661f787e149847348f693986cd49e45

libraries/splunk_app.rb
nmattam: Not necessary 8a626089e661f787e149847348f693986cd49e45

Vagrantfile
nmattam: Updated : ece1b4939290f5f7d5f72b8410d44138e5690236

gravesb:
It would be nice to get some chefspec coverage of this code, but otherwise it looks good.

libraries/splunk_app.rb
```diff
@@ -266,6 +273,18 @@ def install_from_tar(filename, expected_version, installed_version)
         tarfile.close if tarfile
       end
 
+      def manage_lookups
+        lookups = new_resource.lookups
+        lookups.each do |file_name, url|
+          if url.nil?
```
acharlieh: If we change this to `if url && !url.empty?` and flip the conditions, that means we get setting the url to false or empty string for free.

docs/databags.md
```diff
@@ -109,6 +109,7 @@ An apps hash is a contextual (see above) Hash, part of a plaintext data bag item
 * `[app]['download']['version']` - Expected [version number][app.conf] (if any) used to determine if a new app should be downloaded.
 * `[app]['files']` - Hash of files to manage under the "default" or "local" directory.
 * `[app]['files'][filename]` - Contents of a particular file to manage. It can take 3 values, a hash of stanzas -> key-value pairs (then written with the splunk template), a string (written as is), or nil / false (deleted). If the hash or string is empty, the file is also deleted.
+* `[app]['lookups']` - Hash of lookup files for the app. The key in the hash will be the name of file when it lands on splunk and value will be the url to lookup file. To delete an existing lookup file, set the value of the lookup file to `null`. Only supported file name formats are .csv, .gz and .kmz. If you are downloading a comma separated text file (.txt), the file name you specify as the key in the hash should have a .csv extension.
```
acharlieh: We probably should point to Splunk's documentation on the supported file types, that Splunk only supports a plaintext CSV, GZipped CSV, or KMZ for files. Possibly multiple links as CSV is documented separately from KMZ. Could also note that they can use the files hash to maintain any .conf settings that are needed to be set.

libraries/splunk_app.rb
nmattam: Updated here: 754b5a382f2dbe1a5f513216edd31120fa37da4c

Also updated the check for file extension.

docs/databags.md
nmattam: Updated here: 391ed79caf0a1e6a2b30761dc63a68c4486a83b8

nmattam:
Added chefspec test. 5934df97071c841fb4072f4c1431db49481681d7

gravesb:
+1

libraries/splunk_app.rb
```diff
@@ -266,6 +273,18 @@ def install_from_tar(filename, expected_version, installed_version)
         tarfile.close if tarfile
       end
 
+      def manage_lookups
+        lookups = new_resource.lookups
+        lookups.each do |file_name, url|
+          if url && !url.empty?
+            fail "Unsupported lookup file format for #{file_name} in the app #{new_resource.app}" unless file_name =~ /\.(?:csv.gz|csv|kmz)$/i
```
acharlieh: nit: you need to escape the period in `csv.gz`

docs/databags.md
```diff
@@ -109,6 +109,7 @@ An apps hash is a contextual (see above) Hash, part of a plaintext data bag item
...
 * `[app]['files']` - Hash of files to manage under the "default" or "local" directory.
 * `[app]['files'][filename]` - Contents of a particular file to manage. It can take 3 values, a hash of stanzas -> key-value pairs (then written with the splunk template), a string (written as is), or nil / false (deleted). If the hash or string is empty, the file is also deleted.
+* `[app]['lookups']` - Hash of lookup files for the app. The key in the hash will be the name of file when it lands on splunk and value will be the url to lookup file. To delete an existing lookup file, set the value of the lookup file to `false` or `null` or to an empty string. Only supported file name formats are .csv, .csv.gz and .kmz. If you are downloading a comma separated text file (.txt), the file name you specify as the key in the hash should have a .csv extension. Please see [splunk docs](http://docs.splunk.com/Documentation/Splunk/6.3.2/Knowledge/Addfieldsfromexternaldatasources) for more information about the supported file formats. `[app]['files']` hash can be used to specify any .conf setting that is required for the lookup.
```
acharlieh: `when it lands in the Splunk app` ?

`The only supported file name extensions are` ? (if so then the sentence afterward about text file is redundant).

libraries/splunk_app.rb
nmattam: Addressed in this commit: https://github.com/nmattam/cerner_splunk/commit/87fd4fb5ccacd0fdf083317dd0804c8f913e747e

Commit still not showing up in the PR due to some issues with github.

docs/databags.md
nmattam: Updated: https://github.com/nmattam/cerner_splunk/commit/87fd4fb5ccacd0fdf083317dd0804c8f913e747e

acharlieh:
This is merged with 12a41098dd0f2f430ab6132a81c0bd17b348a347 but I'm leaving this open to see if github catches up per [current status](https://status.github.com/)

PR #119, opened 2016-05-19
Allow the splunk.secret file to be configured via vault data bag

gravesb:
This PR allows the splunk.secret file to be configured via chef vault data bag.  It only allows it to be set once though on initial startup.  We'll have to make a second pass through with some further enhancements if we want to support modification of that file on an existing splunk instance.

This is a step toward supporting search head clustering.


recipes/_configure_secret.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  ruby_block 'Check splunk.secret file' do
+    block do
+      CernerSplunk.validate_secret_file(secret_path, secret)
+    end
+  end
+
+  template 'splunk.secret' do
+    path secret_path
+    source 'splunk.secret.erb'
```
acharlieh: Instead of using a template that just writes the contents to a file, we could use just a file resource with the contents as an argument just like we did with the admin password: https://github.com/cerner/cerner_splunk/blob/1.14.1/recipes/_generate_password.rb#L38

recipes/_configure_secret.rb
```diff
@@ -0,0 +1,37 @@
...
+    variables value: secret
+    user node['splunk']['user']
+    group node['splunk']['group']
+    mode '0400'
+    sensitive true
+    notifies :restart, 'service[splunk]'
```
acharlieh: In current state, where we're creating only if it doesn't exist on first install otherwise we're failing, I don't think we need to notify anything... but if we do, I think we should be notifying the marker file immediately (like the other resources on this?)

recipes/_configure_secret.rb
gravesb: Fixed: https://github.com/cerner/cerner_splunk/pull/119/commits/00a387499a84e132af20a1919049635f93bd4d77

recipes/_configure_secret.rb
gravesb: Good point.  Removed notifies here: https://github.com/cerner/cerner_splunk/pull/119/commits/00a387499a84e132af20a1919049635f93bd4d77

acharlieh:
+1

nmattam:
+1

acharlieh:
I was thinking about this over the weekend / this morning, and I think we missed something in the design... 

If a node doesn't need a secret, it shouldn't have to be added to the vault to be able to read the secrets just to find out that it doesn't need one. 

Likewise, groups of nodes shouldn't have to be able to read other secrets that they don't need to be able to put onto themselves.

gravesb:
> If a node doesn't need a secret, it shouldn't have to be added to the vault to be able to read the secrets just to find out that it doesn't need one.

Good catch.  I refactored the code a bit and re-based on the latest stable branch.  Nodes without configured secrets no longer require access to a vault.

acharlieh:
+1

nmattam:
+1 to the updates

PR #121, opened 2016-06-02
Support encrypting passwords when written to conf files

nmattam:
This enhancement will support writing Splunk encrypted passwords in the config files.

## Why ?
Splunk encrypts passwords written in the config files. This causes Chef to detect a password which is different from what was written before causing Splunk to restart (to reflect the change). Writing the encrypted passwords into the conf file will prevent this restart from happening every time.

### DIFF
--- a/libraries/splunk_password.rb
+++ b/libraries/splunk_password.rb
@@ -1,0 +1,46 @@
+# coding: UTF-8
+#
+# Cookbook Name:: cerner_splunk
+# File Name:: splunk_password.rb
+
+require 'base64'
+
+# Module contains different functions to encrypt and decrypt splunk passwords
+module CernerSplunk
+  # Encrypts the password before writing into config files. As of now all the passwords
+  # needs to be XORed except for the sslKeyFilePassword. The boolean
+  # parameter xor controls the XOR logic.
+  def self.splunk_encrypt_password(plain_text, splunk_secret, xor = true)
+    rc4key = splunk_secret.strip[0..15]
+
+    password =
+      if xor
+        pwd = plain_text.unpack('c*')
+        xorkey = get_xor_key(pwd.size)
+        pwd.zip(xorkey).map { |c1, c2| c1 ^ c2 }.pack('c*')
+      end || plain_text
+
+    '$1$' + Base64.encode64(CernerSplunk::RC4.new(rc4key).encrypt("#{password}\0")).strip!
+  end
+
+  # Decrypts the splunk passwords. As of now the encrypted passwords needs to be XORed
+  # to retrieve the plain_text for every password except the sslKeyFilePassword.
+  # The boolean parameter xor controls the XOR logic.
+  def self.splunk_decrypt_password(encryp_password, splunk_secret, xor = true)
+    rc4key = splunk_secret.strip[0..15]
+    pwd = CernerSplunk::RC4.new(rc4key).decrypt(Base64.decode64(encryp_password.sub('$1$', ''))).chomp("\0")
+
+    return pwd unless xor
+
+    password = pwd.unpack('c*')
+    xorkey = get_xor_key(password.size)
+    password.zip(xorkey).map { |c1, c2| c1 ^ c2 }.pack('c*')
+  end
+
+  # Return the key used to XOR with the password
+  def self.get_xor_key(password_size)
+    xorkey = 'DEFAULTSA'.unpack('c*')
+    xorkey += xorkey while xorkey.size < password_size
+    xorkey
+  end
+end
--- a/libraries/splunk_template.rb
+++ b/libraries/splunk_template.rb
@@ -54,6 +54,12 @@ class Chef
         if args || val
           val
         else
+          # evaluating the procs if an attribute in the stanza has a value which is a proc
+          stanzas.each do |stanza, contents|
+            contents.each do |attribute, value|
+              stanzas[stanza][attribute] = value.call if value.is_a? Proc
+            end
+          end
           { stanzas: stanzas }
         end
       end
--- a/recipes/_configure_authentication.rb
+++ b/recipes/_configure_authentication.rb
@@ -108,7 +108,7 @@ when 'LDAP'
     if hash['bindDNpassword']
       password = CernerSplunk::DataBag.load hash['bindDNpassword'], default: default_coords, type: :vault
       fail 'Password must be a String' unless password.is_a?(String)
-      hash['bindDNpassword'] = password
+      hash['bindDNpassword'] = proc { CernerSplunk.splunk_encrypt_password password, node.run_state['cerner_splunk']['splunk.secret'] }
     end
 
     # Verify Attributes
--- a/recipes/_configure_server.rb
+++ b/recipes/_configure_server.rb
@@ -33,6 +33,11 @@ MASTER_ONLY_CONFIGS = %w(
   commit_retry_time
 ).freeze
 
+# default pass4SymmKey value is 'changeme'
+server_stanzas['general']['pass4SymmKey'] = proc { CernerSplunk.splunk_encrypt_password 'changeme', node.run_state['cerner_splunk']['splunk.secret'] }
+# default sslKeysfilePassword value is 'password'
+server_stanzas['sslConfig']['sslKeysfilePassword'] = proc { CernerSplunk.splunk_encrypt_password 'password', node.run_state['cerner_splunk']['splunk.secret'], false }
+
 case node['splunk']['node_type']
 when :search_head, :server
   clusters = CernerSplunk.all_clusters(node).collect do |(cluster, bag)|
@@ -45,7 +50,9 @@ when :search_head, :server
 
     server_stanzas[stanza] = {}
     server_stanzas[stanza]['master_uri'] = master_uri
-    server_stanzas[stanza]['pass4SymmKey'] = pass unless pass.empty?
+
+    server_stanzas[stanza]['pass4SymmKey'] = proc { CernerSplunk.splunk_encrypt_password pass, node.run_state['cerner_splunk']['splunk.secret'] } unless pass.empty?
+
     # Until we support multisite clusters, set multisite explicitly false
     server_stanzas[stanza]['multisite'] = false
     stanza
@@ -159,20 +166,6 @@ server_stanzas['license'] = {
 }
 
 splunk_template 'system/server.conf' do
-  stanzas do
-    old_stanzas = CernerSplunk::Conf::Reader.new("#{node['splunk']['home']}/etc/system/local/server.conf").read
-
-    old_stanzas.each do |key, value|
-      case key
-      when 'general'
-        server_stanzas['general']['guid'] = value['guid'] if value['guid']
-        server_stanzas['general']['pass4SymmKey'] = value['pass4SymmKey'] if value['pass4SymmKey']
-      when 'sslConfig'
-        server_stanzas['sslConfig']['sslKeysfilePassword'] = value['sslKeysfilePassword']
-      end
-    end
-
-    server_stanzas
-  end
+  stanzas server_stanzas
   notifies :restart, 'service[splunk]'
 end
--- a/spec/unit/libraries/splunk_password_spec.rb
+++ b/spec/unit/libraries/splunk_password_spec.rb
@@ -1,0 +1,30 @@
+# coding: UTF-8
+
+require_relative '../spec_helper'
+require 'splunk_password'
+require 'rc4'
+
+describe 'CernerSplunk::splunk_password' do
+  let(:splunk_secret) { 'this_is_the_splunk_secret_key' }
+  let(:password) { 'password' }
+
+  describe '.splunk_decrypt_password' do
+    context 'when decrypting an encrypted password' do
+      subject { CernerSplunk.splunk_encrypt_password(password, splunk_secret) }
+      let(:decrypted_password) { CernerSplunk.splunk_decrypt_password(subject, splunk_secret) }
+
+      it 'results in the original value' do
+        expect(decrypted_password).to eq(password)
+      end
+    end
+
+    context 'when decrypting an encrypted sslconfig password' do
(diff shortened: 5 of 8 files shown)


libraries/splunk_password.rb
```diff
@@ -0,0 +1,99 @@
+# coding: UTF-8
+#
+# Cookbook Name:: cerner_splunk
+# File Name:: splunk_password.rb
+
+require 'base64'
+
+# Module contains different functions to encrypt and decrypt splunk passwords
+module CernerSplunk
+  # RC4 implementation taken from https://github.com/caiges/Ruby-RC4
```
acharlieh: 1) If we're going to copying an implementation, we need to conform to the terms of the license (MIT) in this case. 
2) This is [part of a gem](https://rubygems.org/gems/ruby-rc4) should we be using chef-gem to install and manage the version of it, instead of copying this code into our codebase?

recipes/_configure_authentication.rb
```diff
@@ -108,7 +108,11 @@
     if hash['bindDNpassword']
       password = CernerSplunk::DataBag.load hash['bindDNpassword'], default: default_coords, type: :vault
       fail 'Password must be a String' unless password.is_a?(String)
-      hash['bindDNpassword'] = password
+      ruby_block 'set bindDNpassword' do
+        block do
+          hash['bindDNpassword'] = CernerSplunk.splunk_encrypt_password password, ::File.join(node['splunk']['home'], 'etc/auth/splunk.secret')
```
acharlieh: Should someone invoking the method have to know the location of the splunk secret file or can we infer it?

libraries/splunk_password.rb
nmattam: I prefer to use the class itself over installing the gem so that we can avoid the pain of bringing in an external dependency and managing it. 
I'm not sure about what needs to be done to conform to the MIT license standards. I'm looking into it now.

nmattam: I added the MIT license at the top of the file.

recipes/_configure_authentication.rb
nmattam: Updated: f61854fc62c0eb7fc8b3afa90cd0ac02f3e0b0ca

libraries/splunk_password.rb
nmattam: Charlie suggested moving the RC4 class to a separate file to avoid confusions around Licenses.
51836556ce605c65d15c25f55f9093fbbe972958

recipes/_configure.rb
```diff
@@ -16,6 +16,13 @@
   end
 end
 
+ruby_block 'read splunk.secret' do
```
acharlieh: This should go https://github.com/cerner/cerner_splunk/blob/23a5a65b05e030f6e8d12013e492e6c0c3b60e97/recipes/_install.rb#L81

Immediately after the splunk.secret is either written or generated.

recipes/_configure_server.rb
```diff
@@ -45,7 +55,13 @@
 
     server_stanzas[stanza] = {}
     server_stanzas[stanza]['master_uri'] = master_uri
-    server_stanzas[stanza]['pass4SymmKey'] = pass unless pass.empty?
+
+    ruby_block 'set pass4SymmKey for individual cluster stanza' do
```
acharlieh: You need to include the cluster name in the resource name otherwise resource cloning.

recipes/_configure_server.rb
```diff
@@ -33,6 +36,13 @@
   commit_retry_time
 ).freeze
 
+ruby_block 'set pass4SymmKey and sslKeysfilePassword' do
+  block do
+    server_stanzas['general']['pass4SymmKey'] = CernerSplunk.splunk_encrypt_password default_pass4symmkey, node.run_state['cerner_splunk']['splunk.secret']
```
acharlieh: This block won't execute until converge time, so these need to use `||=` instead of `=`

recipes/_configure_authentication.rb
```diff
@@ -108,7 +108,11 @@
     if hash['bindDNpassword']
       password = CernerSplunk::DataBag.load hash['bindDNpassword'], default: default_coords, type: :vault
       fail 'Password must be a String' unless password.is_a?(String)
-      hash['bindDNpassword'] = password
+      ruby_block 'set bindDNpassword' do
```
acharlieh: Same comment as https://github.com/cerner/cerner_splunk/pull/121/files#r65924306 

Refer to the strategy name in this resource name if we want to do things this way.

Alternatively... should we have an enhancement to the app resource to loop over all of the k-v pairs and evaluate  value lambdas? 

so this becomes 

    hash['bindDNpassword'] = -> { CernerSplunk.splunk_encrypt_password password, node.run_state['cerner_splunk']['splunk.secret'] }

The question becomes if where we would evaluate the lambda, do we have access to the node run state? Could also have a method that produces such lambdas, currying in the password and if the xor needs to be done), but *shrug*

spec/unit/libraries/splunk_password_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+require_relative '../spec_helper'
+require 'splunk_password'
+require 'rc4'
+
+describe 'CernerSplunk::splunk_password' do
+  let(:splunk_secret) { 'this_is_the_splunk_secret_key' }
+  let(:plain_text_password) { 'password' }
+  subject { CernerSplunk.splunk_encrypt_password(plain_text_password, splunk_secret) }
+
+  context 'when string to be encrypted is nil or empty' do
```
jdmurphy: When testing a library with rspec, it is cleaner and easier if you use a `describe` block for each function you want to test. This would look something like 
```
describe 'CernerSplunk::splunk_password' do
  let(:splunk_secret) { 'this_is_the_splunk_secret_key' }
  let(:plain_text_password) { 'password' }
  
  describe '.splunk_encrypt_password' do
    subject { CernerSplunk.splunk_encrypt_password(plain_text_password, splunk_secret) }
    ...
  end
  describe '.splunk_decrypt_password' do
    subject { CernerSplunk.splunk_decrypt_password(plain_text_password, splunk_secret) }
    ...
  end
end
```

recipes/_configure_server.rb
nmattam: Talked to @acharlieh regarding this. Since we are not providing the option to configure the pass4symmkey and sslkeysfilepassword, this won't be a problem.

spec/unit/libraries/splunk_password_spec.rb
nmattam: 3f54b21d9f3885957103fd2caa716ea3ac3b1175

recipes/_configure.rb
nmattam: b2d03f294d61f67d7b109be0cbababf1324e0f35

recipes/_configure_authentication.rb
nmattam: This seems to be doing it : bf65f0a46e05f6f7ad9a155568f6a5820f729f26

libraries/splunk_password.rb
```diff
@@ -0,0 +1,52 @@
...
+
+require 'base64'
+
+# Module contains different functions to encrypt and decrypt splunk passwords
+module CernerSplunk
+  # Encrypts the password before writing into config files. As of now all the passwords
+  # needs to be XORed except for the sslKeyFilePassword. The boolean
+  # parameter xor controls the XOR logic.
+  def self.splunk_encrypt_password(plain_text, splunk_secret, xor = true)
+    fail 'string to be encrypted cannot be empty/nil' if plain_text.nil? || plain_text.empty?
```
acharlieh: I'm curious about why plain_text must not be empty

libraries/splunk_password.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  # Decrypts the splunk passwords. As of now the encrypted passwords needs to be XORed
+  # to retrieve the plain_text for every password except the sslKeyFilePassword.
+  # The boolean parameter xor controls the XOR logic.
+  def self.splunk_decrypt_password(encryp_password, splunk_secret, xor = true)
+    fail 'string to be decrypted cannot be empty/nil' if encryp_password.nil? || encryp_password.empty?
+
+    rc4key = splunk_secret.strip[0..15]
+    pwd = CernerSplunk::RC4.new(rc4key).decrypt(Base64.decode64(encryp_password.gsub('$1$', '')))
```
acharlieh: I don't think you want gsub... rather encrypted form must start with `$1$` and you only want to remove the first instance of this only... Admittedly though $ shouldn't appear elsewhere in a base 64 encoded string, so maybe it's ok.

libraries/splunk_template.rb
```diff
@@ -54,6 +54,12 @@ def variables(args = nil)
         if args || val
           val
         else
+          # evaluating the procs if an attribute in the stanza has a value which is a proc
+          stanzas.each do |stanza, contents|
+            contents.each do |attribute, value|
+              stanzas[stanza][attribute] = value.call if value.class == Proc
```
acharlieh: instead of `value.class == Proc` we probably want value.is_a? Proc

recipes/_configure_server.rb
```diff
@@ -166,9 +174,6 @@
       case key
       when 'general'
         server_stanzas['general']['guid'] = value['guid'] if value['guid']
```
acharlieh: Let's also get rid of this... this was the case with Splunk 4, but ever since Splunk 5.0 and later guid doesn't exist in the server.conf, thus we don't need this delayed evaluation at all (with the addition of value procs :) )

libraries/splunk_password.rb
gravesb: +1 on making decrypt a more exact reversal of encrypt. Also on the line below it could be .chomp("\0") rather than .delete() to only strip that from the end of the password.

recipes/_configure_server.rb
```diff
@@ -159,20 +167,6 @@
 }
 
 splunk_template 'system/server.conf' do
-  stanzas do
-    old_stanzas = CernerSplunk::Conf::Reader.new("#{node['splunk']['home']}/etc/system/local/server.conf").read
-
-    old_stanzas.each do |key, value|
-      case key
-      when 'general'
-        server_stanzas['general']['guid'] = value['guid'] if value['guid']
```
gravesb: do we not need to keep the guid from the old stanzas anymore?

nmattam: https://github.com/cerner/cerner_splunk/pull/121#discussion_r66640786

recipes/_configure_server.rb
```diff
@@ -5,6 +5,9 @@
 #
 # Configures the system server.conf file
 
+default_pass4symmkey = 'changeme'
+default_ssl_password = 'password'
```
gravesb: do these need to be configurable cookbook attributes? Or I suppose that could be a future enhancement to support a non-default ssl keys file/password.

nmattam: Had talked to Charlie regarding this. We decided not to do it now and we will consider doing in the future as this will include more work.

libraries/splunk_password.rb
nmattam: I didn't see a reason why someone should be given the option to encrpyt/decrypt an empty string and hence added this check.

spec/unit/libraries/splunk_password_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+describe 'CernerSplunk::splunk_password' do
+  let(:splunk_secret) { 'this_is_the_splunk_secret_key' }
+  let(:password) { 'password' }
+
+  describe '.splunk_decrypt_password' do
+    context 'when decrypting an encrypted password' do
+      subject { CernerSplunk.splunk_encrypt_password(password, splunk_secret) }
+      let(:decrypted_password) { CernerSplunk.splunk_decrypt_password(subject, splunk_secret) }
+
+      it 'decryted password is same as the plain_text_password' do
```
gravesb: nit, decrypted misspelled, same thing below on line 25 too.

libraries/splunk_password.rb
nmattam: However removed the check here: 5069c7530409bc29b54a3afe750b3dd8e1ef2fb5

libraries/splunk_template.rb
nmattam: 7e1fb3da586474e703e66eaca83f7002bf414886

recipes/_configure_server.rb
nmattam: 081b9e21e4b88772d60440975cac45fc88c864c4

recipes/_configure_server.rb
acharlieh: Future, but admittedly, you use them only in one place, so you could inline them.

spec/unit/libraries/splunk_password_spec.rb
nmattam: 0d5c916840eba969ae2ed84ec76ebd29cec296ed

libraries/splunk_password.rb
nmattam: f8f3c3313a49a924f0f662515798f5a6d32176a9

gravesb:
+1

recipes/_configure_server.rb
nmattam: 1a3e06ac18fe7f4ec5cc70cb7ac5a3e034389252

jdmurphy:
+1

alecxvs:
+1

recipes/_configure_server.rb
```diff
@@ -45,7 +50,9 @@
 
     server_stanzas[stanza] = {}
     server_stanzas[stanza]['master_uri'] = master_uri
-    server_stanzas[stanza]['pass4SymmKey'] = pass unless pass.empty?
+
```
acharlieh: Nit: you added extra spaaaaaaaaaaccccceeee (and line 55)

PR #123, opened 2016-06-16
Allow specification of index size in MB per day

gravesb:
To make the calculation of maxTotalDataSizeMB a little more intuitive, this allows a value in MB per day to be specified in the data bag and the maxTotal index size will be calculated based on that value along with the other cluster size values.

### DIFF
--- a/docs/databags.md
+++ b/docs/databags.md
@@ -18,6 +18,7 @@ The Cluster Hash is part of a plaintext data bag item that defines a logical gro
 * `['master_uri']` - SplunkAPI URI of the cluster master (Required for servers connecting to managed clusters)
 * `['settings']` -  Hash of Cluster settings (Required for servers connecting to managed clusters),
 * `['settings'][???]` - Valid values are those under the clustering stanza of [server.conf][]
+* `['settings'][???]['_cerner_splunk_indexer_count']` - The number of indexers to use for calculating maxTotalDataSizeMB for each index in combination with _maxDailyDataSizeMB in the index configuration.
 * `['replication_ports']` - Configuration for cluster slave replication ports (required for cluster slaves)
 * `['replication_ports']['###']` - Port number to listen on
 * `['replication_ports']['###']['_cerner_splunk_ssl']` - boolean if the port is ssl enabled (false)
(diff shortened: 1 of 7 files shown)


recipes/_configure_indexes.rb
```diff
@@ -36,6 +36,20 @@
       :index
     end
 
+  daily_mb = hash.delete('_maxDailyDataSizeMB')
```
nmattam: Looks like we are assuming that this value should include some padding? If so should the README update say that?

gravesb: I'm not sure I understand your question.  What do you mean it should include padding?

recipes/_configure_indexes.rb
```diff
@@ -36,6 +36,20 @@
...
+  if [:index, :default].include?(stanza_type) && daily_mb && !hash.key?('maxTotalDataSizeMB')
+    settings = CernerSplunk.my_cluster_data(node).fetch('settings', {})
+    replication_factor = settings['replication_factor'] || 1
+    indexer_count = settings['_cerner_splunk_indexer_count'] || 1
+
+    frozen_time_in_days = hash['frozenTimePeriodInSecs'] || config.fetch('default', {})['frozenTimePeriodInSecs']
```
nmattam: Should the variable be _frozen_time_in_secs_ on line 45 and 47 ?

docs/databags.md
```diff
@@ -18,6 +18,7 @@ The Cluster Hash is part of a plaintext data bag item that defines a logical gro
 * `['master_uri']` - SplunkAPI URI of the cluster master (Required for servers connecting to managed clusters)
 * `['settings']` -  Hash of Cluster settings (Required for servers connecting to managed clusters),
 * `['settings'][???]` - Valid values are those under the clustering stanza of [server.conf][]
+* `['settings'][???]['_cerner_splunk_indexer_count']` - The number of indexers to use for calculating maxTotalDataSizeMb for each index in combination with _maxDailyDataSizeMB in the index configuration.
```
nmattam: nit: _maxTotalDataSizeMB_

recipes/_configure_indexes.rb
nmattam: Padding added to ingestion per day to account for the metadata. If the ingestion per day is 50MB and if we decided to add 10% padding, _maxDailyDataSizeMB should be 55MB.

gravesb: Fair point.  Added in a configurable padding here: https://github.com/cerner/cerner_splunk/pull/123/commits/654b2d5bbbcbad2e48d2e172637bf4c1d5cd17f0

recipes/_configure_indexes.rb
gravesb: Sure, changed here: https://github.com/cerner/cerner_splunk/pull/123/files#diff-79be619f4d42bccfc1205531edfcace2R49

docs/databags.md
gravesb: Fixed in https://github.com/cerner/cerner_splunk/pull/123/files#diff-a8cd9604d38ad43b8b92ae521a075a6aR21

jdmurphy:
+1

nmattam:
+1

acharlieh:
+1

PR #124, opened 2016-06-21
Add support for Search Head Clustering

nmattam:
This PR adds support for Search Head Clustering. 
http://docs.splunk.com/Documentation/Splunk/6.4.1/DistSearch/SHCarchitecture


recipes/_update_shcluster.rb
```diff
@@ -0,0 +1,28 @@
...
+elsif node['splunk']['config']['sh_member'] == 'remove'
+  command = 'remove shcluster-member'
+else
+  fail "node['splunk']['config']['sh_member'] accepts only values 'add' and 'remove'" unless node['splunk']['config']['sh_member'].nil?
+end
+
+execute "#{node['splunk']['config']['sh_member']} the search head" do # ~FC037
+  command "cat #{password_file} | xargs -I{} #{node['splunk']['cmd']} #{command} -auth admin:{}"
+  environment 'HOME' => node['splunk']['home']
+  not_if { node['splunk']['config']['sh_member'].nil? }
```
alecxvs: This appears redundant; the run would fail if sh_member is not 'add' or 'remove'.

recipes/_update_shcluster.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+# Recipe determines whether to add/remove this SH member to/from SH Cluster.
+
+password_file = File.join node['splunk']['external_config_directory'], 'password'
+search_heads = CernerSplunk.my_cluster_data(node)['sh_cluster']
+
+fail 'Search Heads are not configured for sh clustering in the cluster databag' if search_heads.nil? || search_heads.empty?
+
+notifies_action = node['splunk']['config']['sh_member'].nil? ? :nothing : :stop
+if node['splunk']['config']['sh_member'] == 'add'
```
alecxvs: nit: We use `node['splunk']['config']['sh_member']` frequently enough that I think it should be assigned a variable.

recipes/_update_shcluster.rb
nmattam: You are right. I can make this a little more neat. Commit to follow.

nmattam: 5feac77124945fdf6ce71e25133123e8b94a7f26

alecxvs:
+1

Vagrantfile
```diff
@@ -3,19 +3,23 @@
...
-  c1_search:    { ip: '33.33.33.10', hostname: 'search.splunk', ports: { 8001 => 8000, 8091 => 8089 } },
   c1_master:    { ip: '33.33.33.11', hostname: 'master.splunk', ports: { 8002 => 8000, 8092 => 8089 } },
   c1_slave1:    { ip: '33.33.33.12', hostname: 'slave01.splunk', ports: { 8003 => 8000, 8093 => 8089 } },
   c1_slave2:    { ip: '33.33.33.13', hostname: 'slave02.splunk', ports: { 8004 => 8000, 8094 => 8089 } },
   c1_slave3:    { ip: '33.33.33.14', hostname: 'slave03.splunk', ports: { 8005 => 8000, 8095 => 8089 } },
-  s_standalone: { ip: '33.33.33.20', hostname: 'splunk2', ports: { 8006 => 8000, 8096 => 8089 } },
-  s_license:    { ip: '33.33.33.30', hostname: 'splunk-license', ports: { 8007 => 8000, 8097 => 8089 } },
+  c1_search1:   { ip: '33.33.33.15', hostname: 'search01.splunk', ports: { 8006 => 8000, 8096 => 8089 } },
+  c1_search2:   { ip: '33.33.33.16', hostname: 'search02.splunk', ports: { 8007 => 8000, 8097 => 8089 } },
+  c1_search3:   { ip: '33.33.33.17', hostname: 'search03.splunk', ports: { 8009 => 8000, 8098 => 8089 } },
```
gravesb: did you intend to skip 8008?

Vagrantfile
```diff
@@ -168,16 +173,55 @@ Vagrant.configure('2') do |config|
...
+  end
+
+  config.vm.define :c1_search3 do |cfg|
+    default_omnibus config
+    cfg.vm.provider :virtualbox do |vb|
+      vb.customize ['modifyvm', :id, '--memory', 256]
+    end
+    cfg.vm.provision :chef_client do |chef|
+      chef_defaults chef, :c1_search3
+      chef.add_role 'sh_captain'
```
gravesb: minor, rather than copying the entire block for 3 and 4, you could make a case statement around line 185 to flex the recipe or role to apply to the node.

docs/databags.md
```diff
@@ -101,7 +104,7 @@ An Alerts hash is a contextual (see above) Hash, part of a plaintext data bag it
 
 Apps Hash
 -----------
-An apps hash is a contextual (see above) Hash, part of a plaintext data bag item or specified directly as attributes that configures apps. A special key of 'master-apps' is looked for managing apps that should be installed and pushed by the cluster master, instead of locally.
+An apps hash is a contextual (see above) Hash, part of a plaintext data bag item or specified directly as attributes that configures apps. A special key of 'master-apps' is looked for managing apps that should be installed and pushed by the cluster master, instead of locally. If search head clustering in enabled, the key 'deployer-apps' is used for managing the apps that should be installed on the deployer and then pushed to the search heads.
```
gravesb: minor, ".. search head clustering __is__ enabled,..."

docs/attributes.md
```diff
@@ -38,6 +38,8 @@ Configurable (with defaults)
 * `node['splunk']['config']['license-pool']` - Data bag item used to configure license pools.
 * `node['splunk']['config']['ui_prefs']` - Hash of stanzas used to configure [ui-prefs.conf][] on the search head in a clustered configuration or a standalone instance.
 * `node['splunk']['config']['assumed_index']` - Name of the index to which data is forwarded to by default, when the index is not configured for the input.(`main`)
+* `node['splunk']['config']['sh_cluster']` - Set this attribute to true to enable SH clustering. (`false`)
+* `node['splunk']['config']['sh_member']` - Specify the action to be performed on the Search Head member. Set this attribute to `add` to add a new SH member to the SH cluster. If set to `remove`, it removes the member from the SH cluster. Add members to the SH cluster using this attribute only if there is aleast one member in the Cluster. Also note that you can only add a clean SH instance to the cluster. (`nil`)
```
jdmurphy: Minor, typo. "only if there is _at least_ one member"

docs/attributes.md
```diff
@@ -38,6 +38,8 @@ Configurable (with defaults)
 * `node['splunk']['config']['license-pool']` - Data bag item used to configure license pools.
 * `node['splunk']['config']['ui_prefs']` - Hash of stanzas used to configure [ui-prefs.conf][] on the search head in a clustered configuration or a standalone instance.
 * `node['splunk']['config']['assumed_index']` - Name of the index to which data is forwarded to by default, when the index is not configured for the input.(`main`)
+* `node['splunk']['config']['sh_cluster']` - Set this attribute to true to enable SH clustering. (`false`)
```
jdmurphy: We should probably spell out Search Head the first time and enclose SH in parenthesis so that it sets the precedent which can be used throughout this document like what you did for the databags markdown file.

recipes/_configure_server.rb
```diff
@@ -41,13 +41,35 @@
...
+
+    if node['splunk']['config']['sh_cluster']
+
+      fail "Deployer URI is not specified for #{cluster}" if deployer_uri.empty?
+
+      replication_ports = bag['replication_ports'] || {}
+      replication_ports.each do |port, port_settings|
+        ssl = port_settings['_cerner_splunk_ssl'] == true
+        stanza = ssl ? "replication_port-ssl://#{port}" : "replication_port://#{port}"
+        server_stanzas[stanza] = port_settings.delete_if do |k, _|
```
gravesb: use reject instead of delete_if so you aren't modifying the underlying configuration hash. I fixed a bunch of these in my last PR (https://github.com/cerner/cerner_splunk/commit/654b2d5bbbcbad2e48d2e172637bf4c1d5cd17f0#diff-099855ec276c01c66dd6f3b1266ef1b0).

recipes/_configure_server.rb
```diff
@@ -41,13 +41,35 @@
...
+          k.start_with? '_cerner_splunk'
+        end
+      end
+
+      server_stanzas['shclustering'] = {}
+      server_stanzas['shclustering']['conf_deploy_fetch_url'] = deployer_uri
+      server_stanzas['shclustering']['disabled'] = 0
+      server_stanzas['shclustering']['replication_factor'] = shc_replication_factor if shc_replication_factor
```
gravesb: This needs to check `shc_replication_factor.empty?`

jdmurphy: I think this will always be true because `if ''` evaluates to true.

recipes/_install.rb
```diff
@@ -37,6 +37,12 @@ def nsp
   notifies :delete, 'file[splunk-marker]', :immediately
 end
 
+service 'splunk-stop' do
+  service_name service
+  action :nothing
+  supports stop: true, nothing: true
```
gravesb: I don't think you need to list nothing here, that's not an actual action you would pass to the service.

recipes/deployer.rb
```diff
@@ -0,0 +1,53 @@
...
+
+attributes = node['splunk']['apps']
+
+attributes_bag = CernerSplunk::DataBag.load(attributes['bag']) || {}
+
+cluster_data = CernerSplunk.my_cluster_data(node) || {}
+
+cluster_bag = CernerSplunk::DataBag.load(cluster_data['apps'], pick_context: ['deployer-apps']) || {}
+
+bag_bag = CernerSplunk::DataBag.load(cluster_bag['bag']) || {}
```
jdmurphy: Variable name could be more descriptive.

spec/unit/recipes/_captain_spec.rb
```diff
@@ -0,0 +1,74 @@
+# coding: UTF-8
```
jdmurphy: This file should just be captain_spec.rb right since the captain recipe doesn't start with an underscore.

recipes/_update_shcluster.rb
```diff
@@ -0,0 +1,29 @@
...
+    notifies_action = :nothing
+    command = "add shcluster-member -current_member_uri #{search_heads.first}"
+  elsif node['splunk']['config']['sh_member'] == 'remove'
+    notifies_action = :stop
+    command = 'remove shcluster-member'
+  else
+    fail "node['splunk']['config']['sh_member'] accepts only values 'add' and 'remove'"
+  end
+
+  execute "#{node['splunk']['config']['sh_member']} the search head" do # ~FC037
```
gravesb: We should probably refactor the code a bit to fix the foodcritic error here rather than ignoring it.

docs/attributes.md
nmattam: 711b71ce20b8d11b2e8b50a5119d76876c55ea03

docs/attributes.md
nmattam: 711b71ce20b8d11b2e8b50a5119d76876c55ea03

recipes/_configure_server.rb
nmattam: 8c8e8253a8a613631d36b0cfec4d76dbdf277f56

recipes/_configure_server.rb
nmattam: Fixed in this commit : 1975a126ed110b99398502dafb844362fa5b7feb
> 
2.1.5 :005 > 2.empty?
NoMethodError: undefined method `empty?' for 2:Fixnum
	from (irb):5

Hence went with 'nil'

spec/unit/recipes/_captain_spec.rb
nmattam: a170b1ee990f8c795b24e1249709fc745cc3533d

recipes/_update_shcluster.rb
nmattam: e7950c36614210cfeffe16f6e4b2e800bafea1b5

spec/unit/recipes/deployer_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+  context 'when sh clustering is disabled' do
+    let(:sh_cluster) { false }
+
+    it 'raises an error' do
+      message = 'SH clustering is disabled. Set node["splunk"]["config"]["sh_cluster"] to true'
+      expect { subject }.to raise_error(RuntimeError, message)
+    end
+  end
+
+  it 'does include cerner_splunk::_install_server' do
```
nmattam: 1975a126ed110b99398502dafb844362fa5b7feb

docs/databags.md
nmattam: 711b71ce20b8d11b2e8b50a5119d76876c55ea03

recipes/_install.rb
nmattam: That was my initial thought. But this was the easiest way to get this done without renaming the 'splunk-start' service.

Now I renamed the 'splunk-start' service to 'splunk' and the old 'splunk' service (used for restart) to 'splunk-restart'. 

b401e22f1ee8a89265e57e19c4ad8c8c5ff6306b

docs/attributes.md
```diff
@@ -36,8 +36,10 @@ Configurable (with defaults)
...
 * `node['splunk']['config']['license-pool']` - Data bag item used to configure license pools.
-* `node['splunk']['config']['ui_prefs']` - Hash of stanzas used to configure [ui-prefs.conf][] on the search head in a clustered configuration or a standalone instance.
+* `node['splunk']['config']['ui_prefs']` - Hash of stanzas used to configure [ui-prefs.conf][] on the search head (SH) in a clustered configuration or a standalone instance.
 * `node['splunk']['config']['assumed_index']` - Name of the index to which data is forwarded to by default, when the index is not configured for the input.(`main`)
+* `node['splunk']['config']['sh_cluster']` - Set this attribute to true to enable search head clustering. (`false`)
+* `node['splunk']['config']['sh_member']` - Specify the action to be performed on the SH member. Set this attribute to `add` to add a new SH member to the SH cluster. If set to `remove`, it removes the member from the SH cluster. Add members to the SH cluster using this attribute only if there is at least one member in the Cluster. Also note that you can only add a clean SH instance to the cluster. (`nil`)
```
gravesb: thoughts on renaming this config to sh_member_action? That would make it a bit more descriptive.

Vagrantfile
nmattam: my bad. fixed here: 84b949e9b5e70c597b24c6b14756895790855155

Vagrantfile
nmattam: 84b949e9b5e70c597b24c6b14756895790855155

recipes/deployer.rb
nmattam: That was copy/pasta from _configure_apps recipe. I renamed it to global_apps_bag because that is what it is used here for.
995d3896515395d882c62fc95a33b659518485bf

docs/attributes.md
nmattam: Agreed. That would make more sense.
940bfee698ca5dfa8361853b174bf4ab6ae9b290

gravesb: Looks like you missed the attributes.md doc here.

nmattam: Forgot Ctrl + S before adding. Updated in the above commit.

gravesb:
+1

recipes/_install.rb
```diff
@@ -29,16 +29,16 @@ def nsp
...
 # Actions
-# This service definition is used only for ensuring splunk is started during the run
-service 'splunk-start' do
+# This service definition is used for ensuring splunk is started during the run and to stop splunk service
+service 'splunk' do
   service_name service
   action :nothing
-  supports status: true, start: true
+  supports status: true, start: true, stop: true
   notifies :delete, 'file[splunk-marker]', :immediately
```
acharlieh: You wouldn't want to delete the restart-needed marker on stop would you? Only on start / restart ?

recipes/_update_shcluster.rb
```diff
@@ -0,0 +1,31 @@
...
+password_file = File.join node['splunk']['external_config_directory'], 'password'
+search_heads = CernerSplunk.my_cluster_data(node)['sh_cluster']
+
+fail 'Search Heads are not configured for sh clustering in the cluster databag' if search_heads.nil? || search_heads.empty?
+
+unless node['splunk']['config']['sh_member_action'].nil?
+  if node['splunk']['config']['sh_member_action'] == 'add'
+
+    execute "#{node['splunk']['config']['sh_member_action']} the search head" do
+      command "cat #{password_file} | xargs -I{} #{node['splunk']['cmd']} add shcluster-member -current_member_uri #{search_heads.first} -auth admin:{}"
```
acharlieh: Instead of propagating knowledge on how we keep track of the admin password around, with the cat -> xargs pattern, should we be pulling the admin password into the node run state? (Then we can use it in lazy attributes), and as long as we update it when we update the admin password we don't care about order as much.

recipes/_update_shcluster.rb
```diff
@@ -0,0 +1,31 @@
...
+# Recipe:: _update_shcluster
+#
+# Recipe determines whether to add/remove this SH member to/from SH Cluster.
+
+password_file = File.join node['splunk']['external_config_directory'], 'password'
+search_heads = CernerSplunk.my_cluster_data(node)['sh_cluster']
+
+fail 'Search Heads are not configured for sh clustering in the cluster databag' if search_heads.nil? || search_heads.empty?
+
+unless node['splunk']['config']['sh_member_action'].nil?
```
acharlieh: This strongly feels like it should be a resource, possibly along with the initialize in the captain recipe... 

action :initialize "Check if I'm a member of a search head cluster, and create with all peers if I am not"
action :add "Check if I am a member of a search head cluster, and add me with one of the peers if I am not"
action :remove "Check if I am a member of a search head cluster, and remove me if I am"

recipes/_install.rb
nmattam: Splunk is stopped after the member is removed from the SHC to avoid messages regarding its removal from appearing on the captain.
There are couple of more commands that need to be executed before you bring that instance up  which we are not supporting now.
http://docs.splunk.com/Documentation/Splunk/6.4.1/DistSearch/Removeaclustermember

I had a splunk-stop service before, which was removed after discussing about it in this [comment](https://github.com/cerner/cerner_splunk/pull/124#discussion_r68251562). I can add it back to make the change passive if @gravesb feels good about it.

gravesb: Actually I think we do want to delete it.  On the off chance that config changed somewhere else on the search head during the run we wouldn't want the delayed service restart action to get kicked off after the service was stopped.

nmattam: I agree that it is a good idea to delete the marker. I was still thinking of a separate resource called splunk-stop (like I had before) so that we can avoid all the renaming.

libraries/matchers.rb
```diff
@@ -1,5 +1,5 @@
 if defined?(ChefSpec)
-  ChefSpec::Runner.define_runner_method :splunk_app
+  ChefSpec.define_matcher :splunk_app
```
nmattam: define_runner_method was deprecated.

recipes/_update_shcluster.rb
nmattam: e8c81e8eccc6304ffa14be74bebc88e0adca7af1

recipes/_update_shcluster.rb
nmattam: 081d4d5f4dd63a24957fc0d14e08b4ca53cf52f9

docs/vagrant.md
```diff
@@ -22,6 +22,13 @@ Running with Vagrant
...
 After you spin up the cluster slaves you will have to restart the cluster master to bring the cluster to a stable state. While spinning up the cluster slaves, they are re-assigned with different GUIDS by the `configure_guids` recipe which requires a restart (this restart can take a while to complete). Cluster master is restarted so that it can identify the new GUIDS.
 
+# Spinning up a Search Head Cluster in Vagrant
+To spin up a SH Cluster, set the `node['splunk']['config']['sh_cluster']` attribute to true and bring up the nodes in the following order using the command below.
+
+<code>vagrant up c1_search1 c1_search2 c1_search3 c1_deployer c1_search4</code>
+
+Once c1_search3 is provisioned (which has the captain role assigned to it), it becomes the captain of the SH Cluster with members c1_search1 and c1_search2. Once the deployer is provisioned, it pushes the apps to cluster memebers. c1_search4 when provisioned joins the cluster.
```
gravesb: minor, members misspelled toward the end here.  "... it pushes the apps to cluster members."

recipes/_install.rb
gravesb: I hadn't considered that before but yes, the service resource renaming is potentially non-passive.  I'm fine reverting most of those changes for the sake of passivity.  

The service resource naming is unfortunate because the splunk-stop resource is going to be largely identical to the splunk-start resource declaration. And when you use it, things will still look redundant since you'll have to call `notifies :stop, 'service[splunk-stop]'`.  Thoughts on reverting the name change but leaving the stop action support here on the splunk-start resource?

spec/unit/resources/sh_cluster.rb
```diff
@@ -0,0 +1,90 @@
...
+    allow(Chef::DataBagItem).to receive(:load).with('cerner_splunk', 'indexes').and_return({})
+    stub_command('/opt/splunk/bin/splunk list shcluster-members -auth admin:admin_password | grep 127.0.0.1').and_return(existing_member)
+    stub_command('/opt/splunk/bin/splunk list shcluster-members -auth admin:admin_password | grep is_captain:1').and_return(existing_member)
+  end
+
+  after do
+    CernerSplunk.reset
+  end
+
+  context 'when the captain needs to be assgined' do
```
gravesb: minor, assigned

spec/unit/resources/sh_cluster.rb
```diff
@@ -0,0 +1,90 @@
...
+    it 'assigns the SH to be the captain' do
+      expect(subject).to run_execute('Captain assignment')
+    end
+  end
+
+  context 'when a new member needs to be added to the cluster and is not an exisiting member of the cluster' do
```
gravesb: existing

spec/unit/resources/sh_cluster.rb
```diff
@@ -0,0 +1,90 @@
...
+  context 'when a member needs to be removed from the cluster and is an existing member of the cluster' do
+    let(:action) { 'remove' }
+    let(:existing_member) { true }
+
+    it 'removes the SH from the SHC' do
+      expect(subject).to run_execute('remove search head')
+    end
+  end
+
+  context 'when a member needs to be removed from the cluster and is an existing member of the cluster' do
```
gravesb: "...and is not an existing member..."

spec/unit/resources/sh_cluster.rb
```diff
@@ -0,0 +1,90 @@
...
+  context 'when a new member needs to be added to the cluster and is not an exisiting member of the cluster' do
+    let(:action) { 'add' }
+    let(:existing_member) { false }
+
+    it 'adds the SH to the SHC' do
+      expect(subject).to run_execute('add search head')
+    end
+  end
+
+  context 'when a new member needs to be added to the cluster and is an exisiting member of the cluster' do
```
gravesb: is *not* an existing member

recipes/_install.rb
nmattam: That was my initial thought. However I didn't feel good about making 'splunk-start' support 'stop' since <code>notifies :stop, 'service[splunk-start]'</code> looks a little confusing and hence went with 'splunk-stop'. We can go that way if everyone feels so.

spec/unit/resources/sh_cluster.rb
nmattam: Line 65 checks for the case when the SH is already a member of the SHC. Line 56 looks for the case when SH is not an existing member.

docs/vagrant.md
nmattam: 3261dbf62f56038a30cd7bac8ec38afc54b9cef2

spec/unit/resources/sh_cluster.rb
nmattam: Made some updates to this test. It should fix it. 61dc0e035f9f931905c14c58001a6f80a94475d9

spec/unit/resources/sh_cluster.rb
nmattam: 61dc0e035f9f931905c14c58001a6f80a94475d9

spec/unit/resources/sh_cluster.rb
gravesb: existing is misspelled

providers/sh_cluster.rb
```diff
@@ -0,0 +1,43 @@
...
+use_inline_resources
+
+action :initialize do
+  search_heads = new_resource.search_heads
+  admin_password = new_resource.admin_password
+
+  execute 'Captain assignment' do
+    command "#{node['splunk']['cmd']} bootstrap shcluster-captain -servers_list '#{search_heads.join(',')}' -auth admin:#{admin_password}"
+    environment 'HOME' => node['splunk']['home']
+    # execute only of there isn't a captain in the cluster
```
jdmurphy: Supremely minor, but you have "of" instead of "if" here.

providers/sh_cluster.rb
```diff
@@ -0,0 +1,43 @@
...
+end
+
+action :add do
+  search_heads = new_resource.search_heads
+  admin_password = new_resource.admin_password
+
+  execute 'add search head' do
+    command "#{node['splunk']['cmd']} add shcluster-member -current_member_uri #{search_heads.first} -auth admin:#{admin_password}"
+    environment 'HOME' => node['splunk']['home']
+    # execute only if this SH is not an exisitng member of the SHC
```
jdmurphy: Minor, but existing is misspelled.

providers/sh_cluster.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+end
+
+action :remove do
+  admin_password = new_resource.admin_password
+
+  execute 'remove search head' do
+    command "#{node['splunk']['cmd']} remove shcluster-member -auth admin:#{admin_password}"
+    environment 'HOME' => node['splunk']['home']
+    # execute only if this SH is an exisitng member of the SHC
```
jdmurphy: [Same comment](https://github.com/cerner/cerner_splunk/pull/124/files#r69913123)

spec/unit/resources/sh_cluster.rb
```diff
@@ -0,0 +1,100 @@
...
+  context 'when the captain needs to be assigned and a captain does not exist in the cluster' do
+    let(:captain_exist) { false }
+    it 'assigns the SH to be the captain' do
+      expect(subject).to run_execute('Captain assignment')
+    end
+  end
+
+  context 'when the captain needs to be assigned and a captain already exist in the cluster' do
+    let(:captain_exist) { true }
+    it 'assigns the SH to be the captain' do
```
jdmurphy: Shouldn't this be `it 'does not assign the SH to be the captain' do`?

nmattam:
Rebased to pull in the character encoding fix. Also squashed.

providers/sh_cluster.rb
nmattam: 92b9d0cf069f8da590b49792ed43a1633eca43d6

providers/sh_cluster.rb
nmattam: 92b9d0cf069f8da590b49792ed43a1633eca43d6

providers/sh_cluster.rb
nmattam: 92b9d0cf069f8da590b49792ed43a1633eca43d6

spec/unit/resources/sh_cluster.rb
nmattam: Yeah. That was copy/paste error
92b9d0cf069f8da590b49792ed43a1633eca43d6

jdmurphy:
+1

docs/recipes.md
```diff
@@ -35,3 +35,19 @@ Installs and configures Splunk Server as a cluster search head. Forwards logs.
...
+
+### `cerner_splunk::shc_captain`
+
+Installs and configures Splunk Server as a search head captain in a search head cluster. Captain assignment initializes the cluster.
+
+### `cerner_splunk::shc_deployer`
+
+Installs and configures Splunk Server as a deployer in a search head cluster. Deploys apps and other configurations to the search heads in a search head cluster.
+
+### `cerner_splunk::server`
```
acharlieh: Duplicates of lines 35-57

docs/vagrant.md
```diff
@@ -22,6 +22,13 @@ Running with Vagrant
 If you want to set up the vagrant cluster to use the license pools defined in the [license pool hash](databags.md#license-pool-hash) databag, add the `configure_guids` recipe to the run_list on the cluster slave (to update the GUIDs on these slaves to predefined values) and update the `license_uri` attribute in the cluster-vagrant databag item to point to the cluster master (_https://33.33.33.30:8089_).
 After you spin up the cluster slaves you will have to restart the cluster master to bring the cluster to a stable state. While spinning up the cluster slaves, they are re-assigned with different GUIDS by the `configure_guids` recipe which requires a restart (this restart can take a while to complete). Cluster master is restarted so that it can identify the new GUIDS.
 
+# Spinning up a Search Head Cluster in Vagrant
+To spin up a SH Cluster, set the `node['splunk']['config']['sh_cluster']` attribute to true and bring up the nodes in the following order using the command below.
```
acharlieh: This isn't still accurate is it?

Also if we name the members right ideally the command is just "vagrant up"

libraries/authenticaton.rb
```diff
@@ -0,0 +1,129 @@
+require_relative 'databag'
+
+# Module for the cookbook
+module CernerSplunk # rubocop:disable ModuleLength
+  ASSUMPTIONS =
```
acharlieh: should we namespace with sub modules?

nmattam:
I made some updates to the branch. There is a separate recipe to remove the search heads from the cluster. shc_search_head recipe by default tries to configure the search head for SHClustering and will add it to the search head. I also made some updates to make the deployer push the authentication.conf, authorize.conf, alerts.conf and outputs.conf files to the search heads.

libraries/authenticaton.rb
jdmurphy: +1 to namespacing

docs/databags.md
```diff
@@ -16,12 +16,13 @@ The Cluster Hash is part of a plaintext data bag item that defines a logical gro
 
 * `['license_uri']` - SplunkAPI URI of the License Server (Required for getting onto the Enterprise license, if unset, use trial license)
 * `['master_uri']` - SplunkAPI URI of the cluster master (Required for servers connecting to managed clusters)
-* `['deployer_uri']` - SplunkAPI URI of the Deployer (Required if `node['splunk']['config']['sh_cluster']` is set to true)
-* `['sh_cluster']` - Array of search head members that are in the SH (Search Head) cluster. Always make sure that the first SH member in the array is an existing member in the cluster. (Required if `node['splunk']['config']['sh_cluster']` is set to true)
+* `['deployer_uri']` - SplunkAPI URI of the Deployer in the Search Head Cluster
+* `['shc_members']` - Array of search head members that are in the SH (Search Head) cluster. Always make sure that the first SH member in the array is an existing member in the cluster.
```
gravesb: Minor, these all have to be IP addresses right? If so, could clarify the doc to state that.

docs/databags.md
```diff
@@ -16,9 +16,13 @@ The Cluster Hash is part of a plaintext data bag item that defines a logical gro
 
 * `['license_uri']` - SplunkAPI URI of the License Server (Required for getting onto the Enterprise license, if unset, use trial license)
 * `['master_uri']` - SplunkAPI URI of the cluster master (Required for servers connecting to managed clusters)
+* `['deployer_uri']` - SplunkAPI URI of the Deployer in the Search Head Cluster
+* `['shc_members']` - Array of search head members that are in the SH (Search Head) cluster. Always make sure that the first SH member in the array is an existing member in the cluster.
```
gravesb: Minor, these all have to be IP addresses right? If so, could clarify the doc to state that.

recipes/_configure_authentication.rb
```diff
@@ -12,133 +12,7 @@
...
-
-    fail "Resolved Role Map for Strategy Name #{strategy_name} is empty!!!" if auth_stanzas["roleMap_#{strategy_name}"].empty?
-    strategy_name
-  end.join(',')
-else
-  fail "Unsupported Auth type '#{hash['authType']}'"
-end
-
 splunk_template 'system/authentication.conf' do
-  sensitive auth_stanzas.any? { |_, v| v.key? 'bindDNpassword' }
```
gravesb: I think we still need to set the sensitive flag here.

libraries/authenticaton.rb
```diff
@@ -0,0 +1,134 @@
+# coding: UTF-8
```
gravesb: The file name is misspelled. should be authentication.rb

providers/sh_cluster.rb
```diff
@@ -0,0 +1,44 @@
...
+
+action :remove do
+  admin_password = new_resource.admin_password
+
+  execute 'remove search head' do
+    command "#{node['splunk']['cmd']} remove shcluster-member -auth admin:#{admin_password}"
+    environment 'HOME' => node['splunk']['home']
+    # execute only if this SH is an existing member of the SHC
+    only_if "#{node['splunk']['cmd']} list shcluster-members -auth admin:#{admin_password} | grep #{node['ipaddress']}"
+    notifies :stop, 'service[splunk]', :immediately
```
acharlieh: This LWRP is set to use inline resources... how are we notifying service[splunk] to stop from here?

providers/sh_cluster.rb
```diff
@@ -0,0 +1,44 @@
...
+# Provider:: sh_cluster
+#
+
+use_inline_resources
+
+action :initialize do
+  search_heads = new_resource.search_heads
+  admin_password = new_resource.admin_password
+
+  execute 'Captain assignment' do
```
acharlieh: Should we be running these as the splunk user as well?

providers/sh_cluster.rb
nmattam: hmmm.. I added it in the first place because foodcritic was complaining the following
>>FC017: LWRP does not notify when updated: ./providers/sh_cluster.rb:9
>>FC017: LWRP does not notify when updated: ./providers/sh_cluster.rb:21
>>FC017: LWRP does not notify when updated: ./providers/sh_cluster.rb:34

and with this in place I can still successfully notify the Splunk service. Not sure how that is happening. I'm looking into it.

providers/sh_cluster.rb
nmattam: Initially I had it. But then noticed that we are not doing it while pushing the [apply-cluster-bundle] (https://github.com/cerner/cerner_splunk/blob/stable/recipes/cluster_master.rb).

recipes/shc_deployer.rb
```diff
@@ -0,0 +1,71 @@
...
+cluster_data = CernerSplunk.my_cluster_data(node) || {}
+
+cluster_bag = CernerSplunk::DataBag.load(cluster_data['apps'], pick_context: ['deployer-apps']) || {}
+
+global_apps_bag = CernerSplunk::DataBag.load(cluster_bag['bag']) || {}
+
+apps = CernerSplunk::SplunkApp.merge_hashes(global_apps_bag, cluster_bag)
+
+# Basic configs for the deployer-configs app
+app_configs = {
```
nmattam: We could have evaluated the contents of the other .conf files here and have created in the app. But proc are not evaluated by the splunk_app resource now. Hence we are writing them to the app later using splunk_template resource.

Vagrantfile
```diff
@@ -3,19 +3,24 @@
...
-  s_standalone: { ip: '33.33.33.20', hostname: 'splunk2', ports: { 8006 => 8000, 8096 => 8089 } },
-  s_license:    { ip: '33.33.33.30', hostname: 'splunk-license', ports: { 8007 => 8000, 8097 => 8089 } },
+  c1_search1:   { ip: '33.33.33.15', hostname: 'search01.splunk', ports: { 8006 => 8000, 8096 => 8089 } },
+  c1_search2:   { ip: '33.33.33.16', hostname: 'search02.splunk', ports: { 8007 => 8000, 8097 => 8089 } },
+  c1_search3:   { ip: '33.33.33.17', hostname: 'search03.splunk', ports: { 8008 => 8000, 8098 => 8089 } },
+  c1_search4:   { ip: '33.33.33.18', hostname: 'search04.splunk', ports: { 8009 => 8000, 8099 => 8089 } },
+  c1_deployer:  { ip: '33.33.33.28', hostname: 'deployer.splunk', ports: { 8010 => 8000, 8100 => 8089 } },
+  s_standalone: { ip: '33.33.33.20', hostname: 'splunk2', ports: { 8011 => 8000, 8101 => 8089 } },
+  s_license:    { ip: '33.33.33.30', hostname: 'splunk-license', ports: { 8012 => 8000, 8102 => 8089 } },
+  s_search:    { ip: '33.33.33.31', hostname: 'splunk-search', ports: { 8013 => 8000, 8103 => 8089 } },
```
nmattam: I didn't have a standalone SH talking to indexer cluster. Adding it back.

recipes/_configure_authentication.rb
nmattam: Fixed in the last commit.

libraries/authenticaton.rb
nmattam: Fixed in the last commit.

libraries/splunk_app.rb
```diff
@@ -69,7 +69,7 @@ def apps_dir(arg = nil)
 
       def local(arg = nil)
         val = set_or_return(:local, arg, kind_of: [TrueClass, FalseClass])
-        (url.nil? || url.empty?) ? val : true
+        url.nil? || url.empty? ? val : true
```
nmattam: Rubocop 0.42.0 was released on July 25th that started complaining about these syntax.

libraries/authenticaton.rb
nmattam: Fixed in my last commit

docs/databags.md
nmattam: Updated 650f737011fb1e329840614bf37fd02b16d26f98

libraries/matchers.rb
```diff
@@ -1,4 +1,6 @@
 if defined?(ChefSpec)
+  ChefSpec.define_matcher :splunk_app
+  ChefSpec.define_matcher :splunk_template
```
gravesb: why were these added? (but not one for cerner_splunk_sh_cluster)

Vagrantfile
gravesb: minor, why change it to s_search and ports 8013,8103?  Since it's part of the (indexer) cluster I think we should keep the c1 prefix but I'm not sure what would make the most sense for the latter part.  Maybe c1_search_standalone? We could also leave it on 8001,8091.

docs/databags.md
```diff
@@ -16,9 +16,13 @@ The Cluster Hash is part of a plaintext data bag item that defines a logical gro
 
 * `['license_uri']` - SplunkAPI URI of the License Server (Required for getting onto the Enterprise license, if unset, use trial license)
 * `['master_uri']` - SplunkAPI URI of the cluster master (Required for servers connecting to managed clusters)
+* `['deployer_uri']` - SplunkAPI URI of the Deployer in the Search Head Cluster
+* `['shc_members']` - Array of Splunk API URIs of the search head members that are in the SH (Search Head) cluster. Always make sure that the first SH member in the array is an existing member in the cluster.
```
gravesb: Minor, I'm guessing the second sentence here could be qualified with "When adding members to an existing cluster, always make sure ..."  Otherwise it would be impossible to setup a new cluster.

Vagrantfile
nmattam: Maybe I'm looking at this in a slightly different way. c1_slave* and c1_master are members of the indexer cluster. c1_search* and c1_deployer are members of a search head cluster. By using "c1_" prefix i'm just grouping them to one cluster. It can either be an index cluster or a SH cluster. s_search is a SH not part of any search head cluster. It is a standalone search head that talks to the indexers in indexer cluster.

The reason why I added s_search was because I wanted a standalone SH standing along with the SHCluster talking to the indexer cluster.  I felt that having the c1_ prefix will cause confusion.

docs/databags.md
```diff
@@ -58,7 +62,7 @@ An Indexes Hash is part of a plaintext data bag item that defines the set of ind
 
 Roles Hash
 ----------
-A Roles Hash is a contextual (see above) Hash, part of a plaintext data bag item that defines roles for every node in a cluster, and is pointed to by the `node[:splunk][:config][:roles]` attribute (usually set in your environment).
+A Roles Hash is a contextual (see above) Hash, part of a plaintext data bag item that defines roles for every node in a cluster, and is pointed to by the `node[:splunk][:config][:roles]` attribute (usually set in your environment). A special key of 'shc_configs' is used for managing the roles on the search heads in a search head cluster.
```
gravesb: minor, All the other keys match roles of nodes (search_head, license_master, etc.) so I think search_head_cluster would make more sense and be more consistent for the special key in all of these contextual hashes.

libraries/splunk_app.rb
acharlieh: Yep both new cops...

providers/sh_cluster.rb
acharlieh: That's because it was written before I knew about run state in chef, so getting the admin password is using a cat... and only root can read that file :)

providers/sh_cluster.rb
acharlieh: As we discussed in person, perhaps the solution here is that the provider doesn't notify from here, but instead the person calling the Resource does the notification (which is enabled by using inline resources).

acharlieh:
Did a bunch of revisions... and squashed commits. If you're interested in my individual commits check out https://github.com/acharlieh/cerner_splunk/compare/11a2450...shc_wip

As I write this I realize I need to address Brett's 3 last comments.

Vagrantfile
```diff
@@ -151,7 +157,7 @@ Vagrant.configure('2') do |config|
   end
 
   # Cruisin' Mos Espa In my Delorean ...
-  (1..3).each do |n|
+  [1].each do |n|
```
gravesb: was this an intentional change? If so, why?

recipes/_configure_server.rb
```diff
@@ -95,12 +96,46 @@
   end
 end
 
+path = "#{node['splunk']['home']}/etc/system/local/server.conf"
+old_stanzas = CernerSplunk::Conf::Reader.new(path).read if File.exist?(path)
```
gravesb: these two lines can move down closer to where they're needed (around line 123).

docs/recipes.md
```diff
@@ -32,6 +32,18 @@ Installs and configures Splunk Server as a cluster slave (indexer). Recieves & I
...
+Installs and configures Splunk Server as a search head in a search head cluster. Forwards logs.
+
+### `cerner_splunk::shc_captain`
+
+Installs and configures Splunk Server as a search head captain in a search head cluster. Captain assignment initializes the search head cluster.
+
+### `cerner_splunk::shc_deployer`
+
+Installs and configures Splunk Server as a deployer in a search head cluster. Deploys apps and other configurations to the search heads in a search head cluster.
+
```
gravesb: Need some doc here for the remove_search_head recipe.

Vagrantfile
acharlieh: I should revert this (and my change to the cluster setup)... I did this so I had an indexer cluster of 1 node in Vagrant so I minimized the number of VMs I needed to have running.

spec/unit/recipes/shc_deployer_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  it 'includes cerner_splunk::_install_server recipe' do
+    expect(subject).to include_recipe('cerner_splunk::_install_server')
+  end
+
+  it 'does not run apply-shcluster-bundle' do
+    expect(subject).not_to run_execute('apply-shcluster-bundle')
+  end
+
+  it 'create the deployer-configs app and notifies the execute[apply-shcluster-bundle] resource to run' do
```
jdmurphy: Super minor, but create -> creates.

docs/databags.md
```diff
@@ -119,6 +124,13 @@ An apps hash is a contextual (see above) Hash, part of a plaintext data bag item
 * `[app]['permissions'][object]['access']['write']` - array of roles or String '*' allowed to write the object
 * `[app]['permissions'][object][???]` - Any other stanza from [default.meta][]
 
+UI Hash
```
nmattam: We don't need this as it is going to be a node attribute. Right?

recipes/cluster_master.rb
```diff
@@ -13,11 +13,10 @@
...
 include_recipe 'cerner_splunk::_install_server'
 
-password_file = File.join node['splunk']['external_config_directory'], 'password'
-
-execute 'apply-cluster-bundle' do
-  command "cat #{password_file} | xargs #{node['splunk']['cmd']} apply cluster-bundle --answer-yes -auth admin:"
+execute 'apply-cluster-bundle' do # ~FC009
+  command lazy { "#{node['splunk']['cmd']} apply cluster-bundle --answer-yes -auth admin:#{node.run_state['cerner_splunk']['admin-password']}" }
   environment 'HOME' => node['splunk']['home']
+  sensitive true
```
nmattam: I have seen this command spitting out something. Are there sensitive contents?

recipes/_configure_shc_alerts.rb
```diff
@@ -0,0 +1,18 @@
...
+# Configures the alert settings for the search heads in a search head cluster
+
+hash = CernerSplunk::DataBag.load node['splunk']['config']['alerts'], pick_context: ['shc_configs']
+
+unless hash
+  Chef::Log.info 'Splunk Alerts not configured for the search heads in the search head cluster.'
+  return
+end
+
+splunk_template 'shcluster/deployer-configs/alert_actions.conf' do
```
nmattam: I remember we had a discussion about changing the name of the app to _shcluster so that 1) its configs will have higher precedence 2) will match the _cluster app that we have in the indexer cluster.

vagrant_repo/environments/splunk_server.json
```diff
@@ -9,7 +9,13 @@
         ],
         "roles": "cerner_splunk/authnz-vagrant:roles",
         "authentication": "cerner_splunk/authnz-vagrant:authn",
-        "alerts": "cerner_splunk/alerts-vagrant:alerts"
+        "alerts": "cerner_splunk/alerts-vagrant:alerts",
+        "secrets": {
+          "c2_boot1": "cerner_splunk/passwords:splunk.secret",
+          "c2_boot2": "cerner_splunk/passwords:splunk.secret",
+          "c2_captain": "cerner_splunk/passwords:splunk.secret",
+          "c2_deployer": "cerner_splunk/passwords:splunk.secret"
```
nmattam: c2_newnode also requires the splunk.secret to decrypt the password even though we are not really testing it in the vagrant set up.

libraries/matchers.rb
acharlieh: It seems that the rspec tests fail without them

recipes/cluster_master.rb
acharlieh: The command itself now contains the admin password for the node. Thus sensitive. 

A future change may be to manage the Splunk session, so we don't have to pass the admin password with every command.

docs/databags.md
acharlieh: Tweaked with bcb7759f22baef53e6a9dc6576bd4c17c678b2f6 will squash in a bit...

docs/recipes.md
acharlieh: Added some with bcb7759f22baef53e6a9dc6576bd4c17c678b2f6 will squash in a bit...

spec/unit/recipes/shc_deployer_spec.rb
acharlieh: Fixed with bcb7759f22baef53e6a9dc6576bd4c17c678b2f6 will squash in a bit...

docs/databags.md
acharlieh: Removed with bcb7759f22baef53e6a9dc6576bd4c17c678b2f6 will squash in a bit...

vagrant_repo/environments/splunk_server.json
acharlieh: c2_newnode gets the splunk.secret when it joins the SH Cluster

acharlieh: will be squashed but: 96c974d5eda5b1b8b1d5ea7cd9245b36265924e1

recipes/_configure_shc_alerts.rb
acharlieh: Yes we did... 8728007071ff1759f8b13bb42d40117ce47ddff7

recipes/_configure_server.rb
acharlieh: While true, I didn't want it inside the if block, as that only applies to SHClusters, but reading a value from existing configuration and pulling it forward is more of a general issue if that makes sense.

docs/databags.md
acharlieh: Used `shcluster` for it... as that's a common abbreviation. 879cf677a96733a571841838761c8768c778afae

Now here's a question... For the apps hash selection should we keep `deployer-apps` or make it match as well? (But also apps for the indexer cluster on the master are pulled with `master-apps`)  (It might be a 🚲🏠💥 but I figure a minor one).

recipes/_configure_server.rb
gravesb: I think I see what you're saying, but right now we only need a value from this file in the case of an SHCluster. Not a big deal, but we're just reading a file unnecessarily for other cases.

acharlieh: That's fair: f74bf07d45902e764cf78bc709c9e49f770c8650

docs/databags.md
```diff
@@ -133,3 +138,4 @@ Docs Navigation
 [alert-actions.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Alert-actionsconf
 [default.meta]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Defaultmetaconf
 [app.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/appconf
+[ui-prefs.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Ui-prefsconf
```
nmattam: minor, not needed anymore.

docs/databags.md
gravesb: I don't have a strong preference either way. deployer-apps seems consistent with master-apps in that it's named for the node that chef will deploy the apps to.

recipes/shc_deployer.rb
```diff
@@ -0,0 +1,71 @@
...
+fail 'Search Heads are not configured for sh clustering in the cluster databag' if search_heads.nil? || search_heads.empty?
+
+instance_exec :shc_deployer, &CernerSplunk::NODE_TYPE
+
+include_recipe 'cerner_splunk::_install_server'
+
+execute 'apply-shcluster-bundle' do
+  command lazy { "#{node['splunk']['cmd']} apply shcluster-bundle -target '#{search_heads.first}' --answer-yes -auth admin:#{node.run_state['cerner_splunk']['admin-password']}" }
+  environment 'HOME' => node['splunk']['home']
+  action :nothing
```
nmattam: sensitive flag ?

gravesb:
+1

nmattam:
+1

jdmurphy:
+1 pending resolution/answering nmattam's last comments.

docs/databags.md
acharlieh: 1f00fb98d00482ce6cb7c5c3fa1a28f174252f20

recipes/shc_deployer.rb
acharlieh: 9b6aa312e053ecc5924efd3a4fc40d4f0b43133c

Issue #125, opened 2016-06-28
Intermittent failures due to wrong ssl config passwords.


nmattam:
While working on https://github.com/cerner/cerner_splunk/pull/124 it was noticed that chef-client run fails intermittently while executing the "change-admin-password" execute resource. Digging deeper into the issue this is because the encrypted form of the sslConfig password written into the server.conf file had the last two characters different from what Splunk encrypted. Since the password changed Splunk encounters issues while restartinng.

splunk_password.rb library encrypted "password" to $1$UM6cTur6RgcA whereas Splunk encrypted it to $1$UM6cTur6RgeT. This mismatch happens only for some splunk.secrets.

For this specific splunk.secret (only the first 16 characters) "qYFEHts8G0E/ABbp" the following test fails.

> Failures:
>   1) CernerSplunk::splunk_password .splunk_decrypt_password when decrypting an encrypted sslconfig password results in the original value
>      Failure/Error: expect(decrypted_password).to eq(password)
> 
>        expected: "password"
>             got: "password\x93"
> 
>        (compared using ==)
>      # ./spec/unit/libraries/splunk_password_spec.rb:26:in `block (4 levels) in <top (required)>'
> 
> > Finished
> 
>  in 9.15 seconds (files took 4.39 seconds to load)
> 2 examples, 1 failure
> Failed examples:

I wasn't able to find the pattern of splunk.secrets that causes this issue. I might have the fix for this issue but waiting for more splunk.secrets to cause this again so that I can confirm my fix works.

If you hit this issue, manually updating the server.conf with the old encrypted password (which you can find from the diff generated during the chef-client run) and restarting Splunk should help.

nmattam:
It seems pass4SymmKey was also causing problems as the captain (in my SHC) was not able to talk to the members (intermittently) since different pass4SymmKey was set on some members.

acharlieh:
Fixed with 6130efd71476070896447fa0513dabf28222a2b7

PR #126, opened 2016-06-30
Fix character encoding issues during RC4 encryption

nmattam:
PR to fix the issue identified in https://github.com/cerner/cerner_splunk/issues/125.

Reading about character encoding, this seems to be the best way to fix issues with any type of character encoding (force_encoding to binary and then use the ASCII vs just taking the bytes). There is also PR open by someone to the Ruby-RC4 project to fix it.

This update fixed the issue with sslConfig and pass4SymmKey passwords that I was experiencing while working on https://github.com/cerner/cerner_splunk/pull/124.

### DIFF
--- a/libraries/rc4.rb
+++ b/libraries/rc4.rb
@@ -38,16 +38,12 @@ module CernerSplunk
     end
 
     def encrypt!(text)
-      index = 0
-      while index < text.length
+      text.force_encoding('binary').unpack('C*').map do |encoded_byte|
         @q1 = (@q1 + 1) % 256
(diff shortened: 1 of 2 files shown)


acharlieh:
+1 ... this is the change in the Pull request that's open against ruby-rc4 yes?

nmattam:
This is the change suggested in one of the PR comment (and also came across this suggestion for fixing encoding issues in few other places). I also did it for the key because that might also cause the same problem.

gravesb:
+1

jdmurphy:
+1

PR #128, opened 2016-08-25
Add proc evaluation to all config file writing

acharlieh:
Also enable attribute reading for doing things like values from a chef-vault and/or encrypting a value to be written to a config file in an app.


docs/databags.md
```diff
@@ -118,13 +118,25 @@ An apps hash is a contextual (see above) Hash, part of a plaintext data bag item
...
 * `[app]['permissions'][object]` - Permissions to manage for a particular knowledge object or class of knowledge objects
 * `[app]['permissions'][object]['access']['read']` - array of roles or String '*' allowed to read the object
 * `[app]['permissions'][object]['access']['write']` - array of roles or String '*' allowed to write the object
 * `[app]['permissions'][object][???]` - Any other stanza from [default.meta][]
 
+Proc Hash
+----------
+In the Apps hash, a proc hash is translated to a Proc to be executed at converge time to determine appropriate value, and parameters for the definition of the same
+
+* `[context]` - the name of the proc to use. Other attributes may be required/optional depending oon this name
```
jdmurphy: Minor, misspelled on as oon.

nmattam:
Since pass4symmkey and sslConfig passwords are encrypted the new way, maybe encrypt the LDAP [auth password](https://github.com/cerner/cerner_splunk/blob/stable/libraries/authentication.rb#L106) in the same way?

docs/databags.md
acharlieh: That was a hiccup anyways. That  whole section shouldn't have been added... fixed.

acharlieh:
@nmattam Good call... added some parameters to the encrypt_splunk method to support this

docs/databags.md
```diff
@@ -118,6 +118,11 @@ An apps hash is a contextual (see above) Hash, part of a plaintext data bag item
...
 * `[app]['files']` - Hash of files to manage under the "default" or "local" directory.
 * `[app]['files'][filename]` - Contents of a particular file to manage. It can take 3 values, a hash of stanzas -> key-value pairs (then written with the splunk template), a string (written as is), or nil / false (deleted). If the hash or string is empty, the file is also deleted.
+* `[app]['files'][filename][stanza][attribute]` - A particular attribute in a configuration file. Can be either a discrete value or a hash (parts described below)
+* `[app]['files'][filename][stanza][attribute]['value']` - Hash that is translated to a proc to determine how to load a value for this attribute at runtime. Required when determining a value through a proc.
+* `[app]['files'][filename][stanza][attribute]['value']['proc']` - Method name from the CernerSplunk::ConfTemplate::Value module to invoke to create the proc. Other keys may be required or optional here (or may come from the general context by default), these map directly to the keyword arguments of the method for generating the proc
```
gravesb: I think this last sentence makes more sense to be on the line above. So:
```
* `[app]['files'][filename][stanza][attribute]['value']` - Hash that is translated to a proc to determine how to load a value for this attribute at runtime. Required when determining a value through a proc.  Certain keys may be required or optional here (or may come from the general context by default), these map directly to the keyword arguments of the method for generating the proc.
* `[app]['files'][filename][stanza][attribute]['value']['proc']` - Method name from the CernerSplunk::ConfTemplate::Value module to invoke to create the proc.
```

same for the transform doc below.

libraries/conf_template.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    # Methods to generate procs for how to transform the value prior to writing.
+    module Transform
+      def self.id(**_)
+        proc { |x| x }
+      end
+
+      def self.splunk_encrypt(node:, xor: true, **_)
+        proc { |x| CernerSplunk.splunk_encrypt_password x, node.run_state['cerner_splunk']['splunk.secret'], xor if x }
```
gravesb: Minor, this would be slightly clearer to read if you included some parens.
`CernerSplunk.splunk_encrypt_password(x, node.run_state['cerner_splunk']['splunk.secret'], xor) if x`

libraries/conf_template.rb
```diff
@@ -0,0 +1,85 @@
+# coding: UTF-8
+#
+# Cookbook Name:: cerner_splunk
+# File Name:: conf_template.rb
+
+module CernerSplunk
+  # Module for generating template data
+  module ConfTemplate
+    # Evaluate a Proc (or a Proc returning Procs)
+    def self.collapse_proc(initial, depth: 10, arguments: {}, label: nil)
```
gravesb: Does the depth ever exceed more than 3-4 at this point?  curious if there was a specific reason for putting it at 10.

acharlieh: Depth doesn't actually ever exceed 1 currently... I wrote this method first, with only vague ideas of how I was going to using it. The thought was there could be situations where we may want to enable procs that return procs which return values and so on, but doing so without a depth option opens the possibility to an infinite loop. `10` was indeed an arbitrary choice, but I have crazy partial thoughts on scripted generation of configuration.

docs/databags.md
acharlieh: Sounds good... Switched.

libraries/conf_template.rb
acharlieh: Added some parenthesis, but obligatory XKCD:

obligatory XKCD is obligatory

nmattam:
+1

gravesb:
+1

jdmurphy:
+1

PR #140, opened 2017-02-17
Splunk 6.5.2 upgrade and point heavy forwarder to license master.

nmattam:
This PR introduces the following changes.

1. Upgrades Splunk to 6.5.2. The upgrade was tested using vagrant. It was noticed during testing that on all instances except license server where master_uri is set to _self_ need to have active_group set to _Trial_ to leverage the enterprise(trial) features. 
2. Updates the _configure_server recipe so that the heavy forwarder now points to the license server so that it can leverage the enterprise features.

### DIFF
--- a/Berksfile
+++ b/Berksfile
@@ -2,4 +2,10 @@ source 'https://supermarket.chef.io'
 
 metadata
 
+case RUBY_VERSION
+when '2.1.6'
+  # build-essential >=8.0.0 requires chef 12.5+
+  cookbook 'build-essential', '< 8.0.0'
+end
+
 cookbook 'cerner_splunk_test', path: 'spec/cookbooks/cerner_splunk_test'
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -37,6 +37,8 @@ fail 'Non-unique hostnames' if @network.collect { |_, v| v[:hostname] }.uniq!
 fail 'Non-unique ports' if @network.collect { |_, v| v[:ports].keys }.flat_map { |v| v }.uniq!
 
 def default_omnibus(config)
+  # https://github.com/cerner/cerner_splunk/issues/141
+  # config.omnibus.chef_version = :latest
   config.omnibus.chef_version = '12.17.44'
 end
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -37,7 +37,7 @@ fail 'Non-unique hostnames' if @network.collect { |_, v| v[:hostname] }.uniq!
 fail 'Non-unique ports' if @network.collect { |_, v| v[:ports].keys }.flat_map { |v| v }.uniq!
 
 def default_omnibus(config)
-  config.omnibus.chef_version = :latest
+  config.omnibus.chef_version = '12.17.44'
 end
 
 def network(config, name, splunk_password = true)
--- a/docs/vagrant.md
+++ b/docs/vagrant.md
@@ -14,8 +14,8 @@ Running with Vagrant
   * Host the root of your mirrored structure on port 8080 using a lightweight HTTP server such as the node package [http-server](https://npmjs.org/package/http-server)
   * Un-comment the `splunk-mirrors` role in the Vagrant file. (Do not check in this modification of your Vagrantfile)
   * Required files and sizes (assuming current cookbook versions)
-    * `splunk-6.3.8-1e8d95973e45-linux-2.6-x86_64.rpm` and `splunk-6.3.8-1e8d95973e45-linux-2.6-amd64.deb` ~ 137MB each
-    * `splunkforwarder-6.3.8-1e8d95973e45-linux-2.6-x86_64.rpm` and `splunkforwarder-6.3.8-1e8d95973e45-linux-2.6-amd64.deb` ~ 16MB each
+    * `splunk-6.5.2-67571ef4b87d-linux-2.6-x86_64.rpm` and `splunk-6.5.2-67571ef4b87d-linux-2.6-amd64.deb` ~ 216MB each
+    * `splunkforwarder-6.5.2-67571ef4b87d-linux-2.6-x86_64.rpm` and `splunkforwarder-6.5.2-67571ef4b87d-linux-2.6-amd64.deb` ~ 19MB each
 * `vagrant-omnibus` installer currently requires internet access to function.
 
 **Note**:
--- a/recipes/_configure_server.rb
+++ b/recipes/_configure_server.rb
@@ -135,13 +135,19 @@ end
 # License Configuration
 license_uri =
   case node['splunk']['node_type']
-  when :forwarder, :license_server
+  when :license_server
     'self'
   when :cluster_master, :cluster_slave, :server, :search_head, :shc_search_head, :shc_captain, :shc_deployer
     if node['splunk']['free_license']
       'self'
     else
       CernerSplunk.my_cluster_data(node)['license_uri'] || 'self'
+    end
+  when :forwarder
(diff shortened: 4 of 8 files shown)


Vagrantfile
```diff
@@ -37,7 +37,7 @@ fail 'Non-unique hostnames' if @network.collect { |_, v| v[:hostname] }.uniq!
 fail 'Non-unique ports' if @network.collect { |_, v| v[:ports].keys }.flat_map { |v| v }.uniq!
 
 def default_omnibus(config)
-  config.omnibus.chef_version = :latest
+  config.omnibus.chef_version = '12.17.44'
```
jdmurphy: Should we also update the Gemfile to be this version?

metadata.rb
```diff
@@ -12,6 +12,8 @@
 depends          'chef-vault', '~> 1.3'
 depends          'ulimit', '~> 0.3'
 depends          'xml', '~> 1.2'
+# build-essential >=8.0.0 requires chef 12.5+
+depends          'build-essential', '< 8.0.0'
```
jdmurphy: This still needed since we are using 12.17.44 in the Vagrantfile and since we should probably update the Gemfile to match?

recipes/_configure_server.rb
```diff
@@ -133,33 +133,49 @@
...
     'self'
   when :cluster_master, :cluster_slave, :server, :search_head, :shc_search_head, :shc_captain, :shc_deployer
     if node['splunk']['free_license']
       'self'
     else
       CernerSplunk.my_cluster_data(node)['license_uri'] || 'self'
     end
+  when :forwarder
+    if node.run_state['cerner_splunk']['heavy_forwarder']
+      CernerSplunk.my_cluster_data(node)['license_uri'] || 'self'
```
acharlieh: I keep thinking about a node attribute here... If set to true -> use cluster data license_uri... otherwise self... 

Having an attribute that has to be set to true for a HWF to use the license_uri, keeps passivity with existing versions, and gives the option for those that need it to set it. 

I also think about the same attribute possibly being able to set an override license_uri... which would enable us to lock down the license master more, and shove all the forwarders off to a different LM (possibly).

Vagrantfile
nmattam: Ran into this [error](https://gist.github.com/nmattam/4660e460eb0f40eb8c43c84c79e239db) while trying to vagrant provision on the stable branch. This error is thrown because _start recipe which creates the init.d is not executed and hence Splunk cannot restart. I was trying to isolate the error and see why _start is being skipped and it seems like the _configure_authentication is raising an exception (although this is not visible in the o/p - even in debug logging) and then it goes to execute all the queued resources (which is trying to restart splunk). If I comment out the _configure_authentication recipe I could successfully provision the nodes.
I locked chef-client to 12.17.44 and everything works fine. Looks like this has to do something with the 12.18.31(latest) version.

nmattam: Looks like we are locking chef to 12.4.3 only for ruby 2.1.6. I didn't run into any issues with the 2.3.1 build which is why I didn't lock the version in the Gemfile.
https://travis-ci.org/cerner/cerner_splunk/builds/202454618

Reason why I locked it here is explained in this [comment](https://github.com/cerner/cerner_splunk/pull/140#discussion_r101807354).

metadata.rb
nmattam: chefspec was failing on ruby 2.1.6 build because of build-essential and reason why I didn't lock it in the Gemfile is mentioned in this [comment](https://github.com/cerner/cerner_splunk/pull/140#discussion_r101816360).

https://travis-ci.org/cerner/cerner_splunk/jobs/202454619

recipes/_configure_server.rb
```diff
@@ -193,7 +209,6 @@
     }
     allotted_pool_size += pool_max_size
   end
-  node.run_state['cerner_splunk'] ||= {}
```
nmattam: I believe having this on line 136 will guarantee it. However, based on other comments in the review, this might end up coming back here.

docs/vagrant.md
```diff
@@ -14,8 +14,8 @@ Running with Vagrant
   * Host the root of your mirrored structure on port 8080 using a lightweight HTTP server such as the node package [http-server](https://npmjs.org/package/http-server)
   * Un-comment the `splunk-mirrors` role in the Vagrant file. (Do not check in this modification of your Vagrantfile)
   * Required files and sizes (assuming current cookbook versions)
-    * `splunk-6.3.8-1e8d95973e45-linux-2.6-x86_64.rpm` and `splunk-6.3.8-1e8d95973e45-linux-2.6-amd64.deb` ~ 137MB each
-    * `splunkforwarder-6.3.8-1e8d95973e45-linux-2.6-x86_64.rpm` and `splunkforwarder-6.3.8-1e8d95973e45-linux-2.6-amd64.deb` ~ 16MB each
+    * `splunk-6.5.2-67571ef4b87d-linux-2.6-x86_64.rpm` and `splunk-6.5.2-67571ef4b87d-linux-2.6-amd64.deb` ~ 137MB each
```
nmattam: e2063cbdc8d7629b07e092b6a23bbc9e60b17b40

metadata.rb
nmattam: cac49aa36b1f9cd12b7f952250238daa90a1ff63

recipes/_configure_server.rb
nmattam: For now, I'm went with the boolean attribute to decide if we want to set to the license_uri or to _self_. 53686564d45fdf5133ee5bd8d780a694d63101d1

recipes/_configure_server.rb
```diff
@@ -133,33 +133,49 @@
...
+  when :license_server
     'self'
   when :cluster_master, :cluster_slave, :server, :search_head, :shc_search_head, :shc_captain, :shc_deployer
     if node['splunk']['free_license']
       'self'
     else
       CernerSplunk.my_cluster_data(node)['license_uri'] || 'self'
     end
+  when :forwarder
+    if node.run_state['cerner_splunk']['heavy_forwarder']
```
nmattam: 53686564d45fdf5133ee5bd8d780a694d63101d1

Vagrantfile
acharlieh: Figured it out #141 If we could have the line with latest commented out with a link to this issue that would be awesome.

nmattam: 809f46c38f5e59c15f3effe8bfd924206ee31dee

recipes/_configure_server.rb
acharlieh: I think this line still needs to be restored.

nmattam: Good catch. I thought I added it back.
a130c949abf8c44e857f763abbc931851068ea9d

Berksfile
```diff
@@ -2,4 +2,10 @@ source 'https://supermarket.chef.io'
 
 metadata
 
+case RUBY_VERSION
+when '2.1.6'
```
alecxvs: I'm concerned about this, how do we know this ruby version corresponds to Chef <= 12.4? Is this assuming ChefDK is being used?

alecxvs:
+1

Berksfile
nmattam: As you [mentioned](https://github.com/cerner/cerner_splunk/pull/140#pullrequestreview-23046625), this chef version (12.4.3) is locked in the Gemfile for this specific ruby version (2.1.6).

jdmurphy:
+1

PR #149, opened 2017-05-04
Update to use sslPassword instead of sslKeysfilePassword

acharlieh:
With the update to 6.5.x, apparently `sslKeysfilePassword` in server.conf was deprecated in favor of `sslPassword`

### DIFF
--- a/libraries/splunk_password.rb
+++ b/libraries/splunk_password.rb
@@ -8,7 +8,7 @@ require 'base64'
 # Module contains different functions to encrypt and decrypt splunk passwords
 module CernerSplunk
   # Encrypts the password before writing into config files. As of now all the passwords
-  # needs to be XORed except for the sslKeyFilePassword. The boolean
+  # needs to be XORed except for the sslKeyFilePassword / sslPassword. The boolean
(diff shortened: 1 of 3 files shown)


libraries/splunk_password.rb
```diff
@@ -8,7 +8,7 @@
 # Module contains different functions to encrypt and decrypt splunk passwords
 module CernerSplunk
   # Encrypts the password before writing into config files. As of now all the passwords
-  # needs to be XORed except for the sslKeyFilePassword. The boolean
+  # needs to be XORed except for the sslKeyFilePassword / sslPassword. The boolean
```
gravesb: minor, should be sslKeysfilePassword (the misspelling kept me from seeing this when I searched through the code). Or just remove it since that attribute is deprecated in splunk.

acharlieh: Huh... how about that, I missed that too. You're right though we should kill references to deprecated settings... We'll throw it in with the next cleanup. I also want to get rid of all the constants that yell when rspec runs: https://github.com/acharlieh/cerner_splunk/tree/sweeping

PR #152, opened 2017-05-12
Splunk 6.6.0

acharlieh:
Also a couple of minor things... removing the mention of a deprecated attribute, and getting rid of a constant... (more cleanup to come later). Suggest minor version increment.


nmattam:
There should be a rspec and README update in all the upgrade PRs. Looks like you missed them.

PR #154, opened 2017-08-17
Remove index validation and cluster data bag validation for forwarders

nmattam:
When we have all the configs (including the outputs.conf) packaged in an app and installed on a fowarder, we don't have to again specify the cluster databag for the forwarder. This PR updates the _configure recipe to log a warning if no cluster data bags are configured rather than throw an error.

@acharlieh also mentioned that we could remove the index check validation in this PR since it doesn't add much value today.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,7 @@ when '2.1.6'
 else
   # https://github.com/cerner/cerner_splunk/issues/142
   chef_version = '= 12.18.31'
+  gem 'molinillo', '~> 0.5.0'
 end
 
 gem 'berkshelf'
--- a/libraries/lwrp.rb
+++ b/libraries/lwrp.rb
@@ -15,7 +15,7 @@ module CernerSplunk
   # Methods involved with augmenting the LWRP syntax / writing recipies
   module LWRP
     # Change a list of monitors to a hash of stanzas for writing to a config file
-    def self.convert_monitors(node, monitors, default_index = nil, base = {})
+    def self.convert_monitors(monitors, default_index = nil, base = {})
       all_stanzas = monitors.inject(base) do |stanzas, element|
         type = element['type'] || element[:type] || 'monitor'
         path = element['path'] || element[:path]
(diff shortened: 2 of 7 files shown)


Gemfile
```diff
@@ -30,6 +30,7 @@ when '2.1.6'
 else
   # https://github.com/cerner/cerner_splunk/issues/142
   chef_version = '= 12.18.31'
+  gem 'molinillo', '~> 0.5.0'
```
nmattam: My tests were failing because build_essentials cookbook was not getting installed by berks. I had run into the same issue before while working on an internal project where I found that the transitive dependencies won't get installed when we have 
_molinillo_ >0.6. Hence locked it to 0.5

recipes/_configure.rb
```diff
@@ -7,7 +7,11 @@
 
 # Verify that clusters are configured
 if node['splunk']['node_type'] != :license_server && node['splunk']['config']['clusters'].empty?
-  throw 'You need to configure at least one cluster databag.'
+  if node['splunk']['node_type'] == :forwarder
+    Chef::Log.warn 'You need to configure at least one cluster databag.'
```
gravesb: The wording here could be better.  Maybe something along the lines of "No cluster data bag configured, ensure your outputs are configured elsewhere."

libraries/lwrp.rb
```diff
@@ -15,7 +15,7 @@ module CernerSplunk
   # Methods involved with augmenting the LWRP syntax / writing recipies
   module LWRP
     # Change a list of monitors to a hash of stanzas for writing to a config file
-    def self.convert_monitors(node, monitors, default_index = nil, base = {})
```
acharlieh: Bikeshedding: on one hand, what if we kept the method signature as is for now just didn't call the other method for passivity, But also this should likely only be an internal detail so, changing the signature and updating everywhere used should still be a non-major release for changing the signature

recipes/_configure.rb
acharlieh: +1

nmattam: dc04e6d08065c81ad7339ca20ad824db1fa50948

nmattam:
Minor version increment 791fff6090086a31cec1e1be2dd6e919b57264bc

PR #157, opened 2017-10-19
Add support for encrypted databags

gravesb:
In order to use this cookbook in an ephemeral environment, like AWS, using chef-vault for secret values becomes extremely difficult.  This adds support for using plain encrypted data bags as an alternative.

I'll squash the commits down to one once the files have been reviewed.


libraries/databag.rb
```diff
@@ -73,27 +76,24 @@ def self.to_value(array, _options = {}) # rubocop:disable CyclomaticComplexity,
...
           begin
-            bag = clazz.load(data_bag, bag_item)
+            bag = case ChefVault::Item.data_bag_item_type(data_bag, bag_item)
+            when :normal
+              data_bag_item(data_bag, bag_item)
+            when :encrypted
+              secret = ::IO.read(opts[:secret]) if opts[:secret]
+              data_bag_item(data_bag, bag_item, secret)
+            when :vault
+              ChefVault::Item.load(data_bag, bag_item)
```
gravesb: I tried to use the chef_vault_item helper method from the chef-vault cookbook here, and it worked fine in the recipe specs and on vagrant.  However, I couldn't get it working in the library rspecs here since it wasn't loading the libraries directory from the chef-vault cookbook.

nmattam:
rubocop didn't like few changes.
https://travis-ci.org/cerner/cerner_splunk/jobs/290010600

gravesb:
Fixed rubocop warnings in https://github.com/cerner/cerner_splunk/pull/157/commits/0e9a79efec109ee7941f1c9d00f723f2f8f839c1.

metadata.rb
```diff
@@ -4,14 +4,14 @@
 license          'Apache-2.0'
 description      'Installs/Configures Splunk Servers and Forwarders'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version          '2.14.0'
```
acharlieh: Typically I defer this change until the merge, but *shrug*

gravesb: I had to do it or berks would fail to resolve the dependencies. Since the previous version is available in the supermarket and I changed the version of chef-vault, it kept telling me I had conflicting dependencies.

gravesb:
As long as you're using the default secret (either in the default location or whatever is configured in the client.rb), you can actually make any of the data bags encrypted with this change.  It's just that you can only configure a separate encryption key for the data bags that we expect to be encrypted.

metadata.rb
acharlieh: Interesting. I wonder if there was a change to Berks, I thought with the metadata directive it was supposed to favor the local metadata.rb.  Although I could see a rogue lock file doing things too... But as I mentioned before *shrug* :)

acharlieh: Interesting. I wonder if there was a change to Berks, I thought with the metadata directive it was supposed to favor the local metadata.rb.  Although I could see a rogue lock file doing things too... But as I mentioned before *shrug* :)

gravesb:
I modified the other recipes to also support the non-default encrypted data bag secret and added a bit to the docs. Take a look again and see what you think.  I figured it was a toss up on whether to have the DataBag helper take the whole node object vs. just the data_bag_secret attribute.

gravesb:
Squashed all commits into one.

PR #158, opened 2017-10-25
Add support for indexer discovery

nmattam:
## What was changed? Why is this necessary?
This PR introduces support for indexer discovery.  http://docs.splunk.com/Documentation/Splunk/6.6.0/Indexer/indexerdiscovery

## How was it tested?
There are ChefSpecs included in this PR. Also, the vagrant set up is now updated so that the forwarder talks to the peers in the clustered setup using indexer discovery and to the standalone instance in the traditional way (where we list the peer IPs using the `servers` attribute in the `tcpout` stanza).

### DIFF
--- a/docs/databags.md
+++ b/docs/databags.md
@@ -30,6 +30,14 @@ The Cluster Hash is part of a plaintext data bag item that defines a logical gro
 * `['shc_replication_ports']` - identical as `['replication_ports']` but take precedence for Search Head Cluster members/captain when replication port settings need to differ between SHC members and indexer cluster slaves in the same cluster.
 * `['receivers']` - Array of strings of hosts where this cluster's indexers are listening. (Required for forwarders)
 * `['receiver_settings']['splunktcp']['port']` - Port indexers are listening on, and forwarders are sending data (required for forwarders and receivers)
+* `['tcpout_settings']` - Hash of tcpout settings to be configured in the outputs.conf under the `tcpout` stanza.
+* `['tcpout_settings'][???]` - Valid values are those under the tcpout stanza of [outputs.conf][]
+* `['indexer_discovery']` - Boolean to toggle indexer_discovery. Set this to `true` to leverage indexer discovery. Note that `['receivers']` configured when `['indexer_discovery']` is set to `true` will be ignored.
+* `['indexer_discovery_settings']['pass4SymmKey']` - pass4Symmkey value to be set under the `index_discovery` stanza in [server.conf][] and [outputs.conf][]
+* `['indexer_discovery_settings']['master_configs']` - Hash of indexer discovery settings to be configured on the cluster master under the `indexer_discovery` stanza.
+* `['indexer_discovery_settings']['master_configs'][???]` - Valid values are those under the indexer_discovery stanza of [server.conf][]
+* `['indexer_discovery_settings']['outputs_configs']` - Hash of indexer discovery settings to be configured in the outputs.conf under the `indexer_discovery` stanza.
+* `['indexer_discovery_settings']['outputs_configs'][???]` - Valid values are those under the indexer_discovery stanza of [outputs.conf][]
 * `['indexes']` - A String pointing to an indexes data bag hash. (Coordinate form as described above)
 * `['apps']` - A String pointing to an apps data bag hash. (Coordinate form as described above)
 
@@ -135,11 +143,12 @@ Docs Navigation
 * [Docs Readme](README.md)
 * [Repository Readme](../README.md)
 
-[server.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Serverconf
-[indexes.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Indexesconf
-[user-prefs.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/User-prefsconf
-[authorize.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Authorizeconf
-[authentication.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Authenticationconf
 [alert-actions.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Alert-actionsconf
-[default.meta]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Defaultmetaconf
 [app.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/appconf
+[authentication.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Authenticationconf
+[authorize.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Authorizeconf
+[default.meta]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Defaultmetaconf
+[indexes.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Indexesconf
+[outputs.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Outputsconf
+[server.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/Serverconf
+[user-prefs.conf]: http://docs.splunk.com/Documentation/Splunk/latest/Admin/User-prefsconf
--- a/libraries/outputs.rb
+++ b/libraries/outputs.rb
@@ -8,7 +8,7 @@ require_relative 'databag'
 module CernerSplunk
   # Module contains functions to configure outputs.conf in a Splunk system
   module Outputs
-    def self.configure_outputs(node) # rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
+    def self.configure_outputs(node) # rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity, Metrics/MethodLength
       output_stanzas = {}
 
       if %i[search_head forwarder cluster_master shc_deployer].include? node['splunk']['node_type']
@@ -24,6 +24,31 @@ module CernerSplunk
         else
           [CernerSplunk.my_cluster(node)]
         end.each do |(cluster, bag)|
+          if bag['indexer_discovery'] == true
+            Chef::Log.warn "Configured ['receivers'] in cluster #{cluster} will be ignored since ['indexer_discovery'] is set to true." if bag['receivers']
+
+            indexer_discovery_settings = ((bag['indexer_discovery_settings'] && bag['indexer_discovery_settings']['outputs_configs']) || {}).reject do |k, _|
+              k.start_with?('_cerner_splunk')
+            end
+            output_stanzas["indexer_discovery:#{cluster}"] = indexer_discovery_settings
+
+            fail "master_uri is missing in the cluster databag: #{cluster}" if bag['master_uri'].nil? || bag['master_uri'].empty?
+
+            output_stanzas["indexer_discovery:#{cluster}"]['master_uri'] = bag['master_uri']
+            encrypt_password = CernerSplunk::ConfTemplate::Transform.splunk_encrypt node: node
+
+            pass =
+              if bag['indexer_discovery_settings'] && bag['indexer_discovery_settings']['pass4SymmKey']
+                bag['indexer_discovery_settings']['pass4SymmKey']
+              else
+                'changeme'
+              end
+            output_stanzas["indexer_discovery:#{cluster}"]['pass4SymmKey'] = CernerSplunk::ConfTemplate.compose encrypt_password, CernerSplunk::ConfTemplate::Value.constant(value: pass)
+            output_stanzas["tcpout:#{cluster}"] = bag['tcpout_settings'] || {}
+            output_stanzas["tcpout:#{cluster}"]['indexerDiscovery'] = cluster
+            next
+          end
+
           port = bag['receiver_settings']
           port = port['splunktcp'] if port
           port = port['port'] if port
@@ -32,7 +57,7 @@ module CernerSplunk
           if !receivers || receivers.empty? || !port
             Chef::Log.warn "Receiver settings missing or incomplete in configured cluster data bag: #{cluster}"
           else
-            output_stanzas["tcpout:#{cluster}"] = {}
+            output_stanzas["tcpout:#{cluster}"] = bag['tcpout_settings'] || {}
             output_stanzas["tcpout:#{cluster}"]['server'] = receivers.collect do |x|
               x.include?(':') ? x : "#{x}:#{port}"
             end.join(',')
--- a/recipes/_configure_inputs.rb
+++ b/recipes/_configure_inputs.rb
@@ -11,15 +11,14 @@ base_hash = { 'default' => { 'host' => node['splunk']['config']['host'] } }
 input_stanzas = CernerSplunk::LWRP.convert_monitors node['splunk']['monitors'], node['splunk']['main_project_index'], base_hash
 
 if %i[server cluster_slave].include? node['splunk']['node_type']
-  bag = CernerSplunk.my_cluster_data(node)
+  cluster, bag = CernerSplunk.my_cluster(node)
   port = bag['receiver_settings']
   port = port['splunktcp'] if port
   port = port['port'] if port
-
   if port
-    input_stanzas["splunktcp://:#{port}"] = {}
-    input_stanzas["splunktcp://:#{port}"]['disabled'] = 0
-    input_stanzas["splunktcp://:#{port}"]['connection_host'] = 'none'
+    input_stanzas["splunktcp://#{port}"] = {}
+    input_stanzas["splunktcp://#{port}"]['disabled'] = 0
+    input_stanzas["splunktcp://#{port}"]['connection_host'] = 'none'
   else
     Chef::Log.warn "Receiver settings missing in configured cluster data bag: #{cluster}"
   end
--- a/recipes/_configure_server.rb
+++ b/recipes/_configure_server.rb
@@ -75,6 +75,17 @@ when :cluster_master
 
   server_stanzas['clustering'] = settings
   server_stanzas['clustering']['mode'] = 'master'
+
+  if bag['indexer_discovery'] == true
+    indexer_discovery_settings = ((bag['indexer_discovery_settings'] && bag['indexer_discovery_settings']['master_configs']) || {}).reject do |k, _|
+      k.start_with?('_cerner_splunk')
+    end
+    pass = (bag['indexer_discovery_settings'] && bag['indexer_discovery_settings']['pass4SymmKey']) || nil
+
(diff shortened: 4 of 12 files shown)


recipes/_configure_inputs.rb
```diff
@@ -11,15 +11,14 @@
...
+  cluster, bag = CernerSplunk.my_cluster(node)
   port = bag['receiver_settings']
   port = port['splunktcp'] if port
   port = port['port'] if port
-
   if port
-    input_stanzas["splunktcp://:#{port}"] = {}
-    input_stanzas["splunktcp://:#{port}"]['disabled'] = 0
-    input_stanzas["splunktcp://:#{port}"]['connection_host'] = 'none'
+    input_stanzas["splunktcp://#{port}"] = {}
```
nmattam: I ran into this issue https://answers.splunk.com/answers/341027/why-am-i-unable-to-connect-my-forwarder-to-an-inde.html and the fix suggested by MuS fixed it.

libraries/outputs.rb
```diff
@@ -24,6 +24,27 @@ def self.configure_outputs(node) # rubocop:disable Metrics/PerceivedComplexity,
...
+            encrypt_password = CernerSplunk::ConfTemplate::Transform.splunk_encrypt node: node
+            pass =
+              if bag['indexer_discovery_settings'] && bag['indexer_discovery_settings']['pass4SymmKey']
+                bag['indexer_discovery_settings']['pass4SymmKey']
+              else
+                'changeme'
+              end
+            output_stanzas["indexer_discovery:#{cluster}"]['pass4SymmKey'] = CernerSplunk::ConfTemplate.compose encrypt_password, CernerSplunk::ConfTemplate::Value.constant(value: pass)
+            output_stanzas["tcpout:#{cluster}"] = {}
+            output_stanzas["tcpout:#{cluster}"]['indexerDiscovery'] = cluster
```
gravesb: Doesn't line 56 below remove this from the output stanzas? Seems like it would need to be updated to `|= {}`.  Is it even valid to list both servers and indexer_discovery in the same tcpout stanza?  We may need some additional error checking to say a given cluster can only configure one of receivers or indexer_discovery. And if indexer discovery is on, then don't log the warning about an incomplete cluster data bag on line 54. Another alternative would be to document that if both are present then the configured receivers take precedence over indexer discovery, though I think I would prefer to error in that case.

libraries/outputs.rb
```diff
@@ -24,6 +24,27 @@ def self.configure_outputs(node) # rubocop:disable Metrics/PerceivedComplexity,
...
+              output_stanzas["indexer_discovery:#{cluster}"]['master_uri'] = bag['master_uri']
+            end
+            encrypt_password = CernerSplunk::ConfTemplate::Transform.splunk_encrypt node: node
+            pass =
+              if bag['indexer_discovery_settings'] && bag['indexer_discovery_settings']['pass4SymmKey']
+                bag['indexer_discovery_settings']['pass4SymmKey']
+              else
+                'changeme'
+              end
+            output_stanzas["indexer_discovery:#{cluster}"]['pass4SymmKey'] = CernerSplunk::ConfTemplate.compose encrypt_password, CernerSplunk::ConfTemplate::Value.constant(value: pass)
```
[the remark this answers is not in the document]
nmattam: > `[indexer_discovery_settings][???]` - Valid values are those under the indexer_discovery stanza of [server.conf][]

README update says these are indexer discovery settings for server.conf. send_timeout, rcv_timeout, cxn_timeout are optional parameters that go to ouputs.conf. pass4Symmkey is what needs to be configured on both outputs.conf and server.conf and its value has to be the same.

libraries/outputs.rb
nmattam: My plan was that indexer discovery if configured takes precedence over the receivers configured in the cluster databag. I have a `next` on line 45 which should skip the receiver configs if indexer discovery is configured.

libraries/outputs.rb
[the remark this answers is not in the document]
gravesb: apparently I didn't read the doc update close enough.  Although I think the question still remains on whether we should allow the optional outputs attributes to be configured through the cookbook. Given that we allow the optional server conf attributes, it feels more consistent to allow the optional attributes here as well.

libraries/outputs.rb
gravesb: overlooked the `next` in there.  I think we need to document that behavior somewhere and in addition either fail the chef run or log a warning saying the receivers will be ignored if both are configured. I would prefer to fail the run.

libraries/outputs.rb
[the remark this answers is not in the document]
nmattam: > Although I think the question still remains on whether we should allow the optional outputs attributes to be configured through the cookbook. Given that we allow the optional server conf attributes, it feels more consistent to allow the optional attributes here as well.

I did think about giving the option to configure those optional attributes. But again it was consistency why I decided not to do that. I don't think that today we are providing this option in outputs.conf. For example, there are multiple optional attributes that can be configured under `[tcpout:]` stanza, but we only support the `servers` attribute. 
As you mentioned we do allow configuring optional attributes for different stanzas in server.conf which is why I decided to provide that option https://github.com/nmattam/cerner_splunk/blob/support-indexer-discovery/recipes/_configure_server.rb#L84 in server.conf.

That being said I'm open to supporting those attributes in outputs.conf. I don't have a strong reason why we shouldn't.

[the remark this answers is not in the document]
nmattam: c0999811ba5186808d0c516dfdfa6c6df0218db9

libraries/outputs.rb
nmattam: Talked to @gravesb about this. Personally, I don't think there is a reason to fail the run here because having the receivers configured is not going to cause any issues. indexer discovery will take precedence and everything should work as expected. I believe logging a warning is sufficient here and that way it will be consistent with how the cookbook handles similar cases. For example, when `receivers` attribute is not configured, we are only logging a warning. https://github.com/cerner/cerner_splunk/blob/stable/libraries/outputs.rb#L33.

For now, I made the update so that it logs a warning. I have also made the README update to mention that receivers configured in the data bag will be ignored if indexer discovery is enabled. 24053d683e023aa69c5b514f64a8b6cc6af294f2

recipes/_configure_inputs.rb
```diff
@@ -11,15 +11,14 @@
 input_stanzas = CernerSplunk::LWRP.convert_monitors node['splunk']['monitors'], node['splunk']['main_project_index'], base_hash
 
 if %i[server cluster_slave].include? node['splunk']['node_type']
-  bag = CernerSplunk.my_cluster_data(node)
+  cluster, bag = CernerSplunk.my_cluster(node)
```
acharlieh: Is it me, or is the introduced cluster variable not used?

nmattam: It is on line 23

acharlieh: Aargh... who wrote this crap... oh yeah :bear:

libraries/outputs.rb
```diff
@@ -24,6 +24,35 @@ def self.configure_outputs(node) # rubocop:disable Metrics/PerceivedComplexity,
         else
           [CernerSplunk.my_cluster(node)]
         end.each do |(cluster, bag)|
+          if bag['indexer_discovery'] == true
+            Chef::Log.warn "Found ['receivers'] configured in cluster #{cluster} when ['indexer_discovery'] is set to true." if bag['receivers']
```
gravesb: minor, the wording here could be a bit more clear.  Something along the lines of "Configured ['receivers'] in cluster #{cluster} will be ignored since ['indexer_discovery'] is set to true."

nmattam: b88a9564d9fad8296edff5790f0db68ee493299e

libraries/outputs.rb
```diff
@@ -24,6 +24,35 @@ def self.configure_outputs(node) # rubocop:disable Metrics/PerceivedComplexity,
...
+          if bag['indexer_discovery'] == true
+            Chef::Log.warn "Configured ['receivers'] in cluster #{cluster} will be ignored since ['indexer_discovery'] is set to true." if bag['receivers']
+
+            indexer_discovery_settings = ((bag['indexer_discovery_settings'] && bag['indexer_discovery_settings']['outputs_configs']) || {}).reject do |k, _|
+              k.start_with?('_cerner_splunk')
+            end
+            output_stanzas["indexer_discovery:#{cluster}"] = indexer_discovery_settings
+
+            if bag['master_uri'].nil? || bag['master_uri'].empty?
+              Chef::Log.warn "master_uri is missing in the cluster databag: #{cluster}"
```
acharlieh: I think this would be a failure state? we would be required to use a master_uri for indexer discovery yes?

libraries/outputs.rb
```diff
@@ -24,6 +24,35 @@ def self.configure_outputs(node) # rubocop:disable Metrics/PerceivedComplexity,
         else
           [CernerSplunk.my_cluster(node)]
         end.each do |(cluster, bag)|
+          if bag['indexer_discovery'] == true
+            Chef::Log.warn "Configured ['receivers'] in cluster #{cluster} will be ignored since ['indexer_discovery'] is set to true." if bag['receivers']
+
+            indexer_discovery_settings = ((bag['indexer_discovery_settings'] && bag['indexer_discovery_settings']['outputs_configs']) || {}).reject do |k, _|
+              k.start_with?('_cerner_splunk')
+            end
+            output_stanzas["indexer_discovery:#{cluster}"] = indexer_discovery_settings
```
acharlieh: I've been thinking about this, and I wonder if we need to run the various settings that we inject through the similar process we use for inserting procs to create values in conf files.

libraries/outputs.rb
```diff
@@ -24,6 +24,35 @@ def self.configure_outputs(node) # rubocop:disable Metrics/PerceivedComplexity,
...
+            if bag['master_uri'].nil? || bag['master_uri'].empty?
+              Chef::Log.warn "master_uri is missing in the cluster databag: #{cluster}"
+            else
+              output_stanzas["indexer_discovery:#{cluster}"]['master_uri'] = bag['master_uri']
+            end
+            encrypt_password = CernerSplunk::ConfTemplate::Transform.splunk_encrypt node: node
+
+            pass =
+              if bag['indexer_discovery_settings'] && bag['indexer_discovery_settings']['pass4SymmKey']
+                bag['indexer_discovery_settings']['pass4SymmKey']
```
acharlieh: is it me, or is this a symmetric key guaranteed to be plaintext ?

libraries/outputs.rb
```diff
@@ -24,6 +24,35 @@ def self.configure_outputs(node) # rubocop:disable Metrics/PerceivedComplexity,
...
+              else
+                'changeme'
+              end
+
+            output_stanzas["indexer_discovery:#{cluster}"]['pass4SymmKey'] = CernerSplunk::ConfTemplate.compose encrypt_password, CernerSplunk::ConfTemplate::Value.constant(value: pass)
+            output_stanzas["tcpout:#{cluster}"] = {}
+            output_stanzas["tcpout:#{cluster}"]['indexerDiscovery'] = cluster
+            output_stanzas["tcpout:#{cluster}"]['useACK'] = true
```
acharlieh: I suspect this should be configurable... the tradeoff is a possibility of duplicated data versus a possibility of lost data from a forwarder.

recipes/_configure_server.rb
```diff
@@ -75,6 +75,15 @@
...
+
+  indexer_discovery_settings = ((bag['indexer_discovery_settings'] && bag['indexer_discovery_settings']['master_configs']) || {}).reject do |k, _|
+    k.start_with?('_cerner_splunk')
+  end
+  pass = (bag['indexer_discovery_settings'] && bag['indexer_discovery_settings']['pass4SymmKey']) || nil
+
+  server_stanzas['indexer_discovery'] = indexer_discovery_settings if bag['indexer_discovery_settings']
+  server_stanzas['indexer_discovery']['pass4SymmKey'] = CernerSplunk::ConfTemplate.compose encrypt_password, CernerSplunk::ConfTemplate::Value.constant(value: pass) if pass
+
 when :cluster_slave
```
acharlieh: For Cluster Slave, I think there are some additional configuration options we may want to support... namely `register_forwarder_address` under the `[clustering]` stanza... there are other interesting settings too like `register_replication_address` and `register_search_address` which would likely be set via procs if necessary so would need that sort of discovery and injection

libraries/outputs.rb
nmattam: Not setting the master_uri would result in indexer discovery not working and hence logs not getting forwarded. I thought this should be handled in the same way when `receivers` are not set. https://github.com/cerner/cerner_splunk/blob/stable/libraries/outputs.rb#L33
That was why I decided to WARN and not to FAIL.

libraries/outputs.rb
nmattam: > is this a symmetric key guaranteed to be plaintext on the chef server

If I understand the question, you are asking if we are going to have the pass4SymmKey as plain text in our data bags on the Chef server. I think that is how we have been dealing with other pass4Symmkeys (that we set in server.conf). Also based on the proposal made on an internal JIRA issue (which is in progress) investigating how to manage pass4Symmkey, the plan is to not encrypt pass4Symmkey on the forwarder because if we need to encrypt, we will either need to share the shared encryption key with all the forwarders or we will need to control the splunk.secret on the forwarders.

libraries/outputs.rb
nmattam: I hardcoded it because I remember that I couldn't get the indexer discovery to work during vagrant testing without this attribute set.

libraries/outputs.rb
acharlieh: Hrm... yet again... who wrote this thing? Yep, I'm a clown... that apparently has been in there since before the open source version... possibly all the way back to the beginning of the project.

There is a slight difference, as here we're in a place where someone has explicitly said they want to use indexer discovery, but failed to provide the required parameter to use it... versus with the receiver bags, we're in a place where we have implied that someone should have configured something... (but also that place doesn't really make much sense to not having receivers configured there either).

nmattam: Updated to throw error and added some tests: b994c7015ea7ea3bec8b39a9d81a6e95218c45de

From our conversation yesterday, I updated the receivers check as well to throw an error instead of a warning. This will be kind of a non-passive change for those who haven't configured the receivers. But based on what we discussed yesterday that is fine because they will now configure receivers which they should.

libraries/outputs.rb
nmattam: Tested this again today morning and it worked without setting useACK. It didn't work before because the pass4symmkey I configured was different on the cluster master and the forwarder (hence I removed the configured pass4symmkey from the databag). Added the option to configure tcpout attributes a07479b681b8d30108c0393832ff5eea872339e8

recipes/_configure_server.rb
nmattam: Talked to Charlie about this yesterday. 
For now, we will move forward without configuring this attribute. Maybe once we are at a point where we can deploy this in AWS and test it, we will consider adding these attributes if required. Charlie suggested using the Instance Metadata (http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) to find the instance public ip if there is one.

libraries/outputs.rb
nmattam: Above commit started failing rspec on ruby 2.1.6. Rubocop didn't like the updates that I made to fix those tests. cc8505883ed9887bec285f2f1acdd4b62fb5da64 should make rubocop and rspec happy.

libraries/outputs.rb
nmattam: For now, I think we can leave it like this. As mentioned in this comment (https://github.com/cerner/cerner_splunk/pull/158#discussion_r148015697) I believe the plan is to move forward with plain text pass4SymmKey.

libraries/outputs.rb
```diff
@@ -24,19 +24,42 @@ def self.configure_outputs(node) # rubocop:disable Metrics/PerceivedComplexity,
...
 
-          if !receivers || receivers.empty? || !port
-            Chef::Log.warn "Receiver settings missing or incomplete in configured cluster data bag: #{cluster}"
-          else
-            output_stanzas["tcpout:#{cluster}"] = {}
-            output_stanzas["tcpout:#{cluster}"]['server'] = receivers.collect do |x|
-              x.include?(':') ? x : "#{x}:#{port}"
-            end.join(',')
-          end
+          fail "Receiver settings missing or incomplete in configured cluster data bag: #{cluster}" if !receivers || receivers.empty? || !port
```
acharlieh: Being a more non-passive change, let's keep the warning for the receivers being missing for now. I'm picturing a use case, where I as an owner of forwarders define a "cluster" as my cluster to define sets of apps that should be on my servers, but use additional "clusters" to define where the logs should go.

But let's do keep the failure for if I've configured indexer discovery and do not specify the master URI.

nmattam: Updated here: 241d83680bfa98a7da6074c0b51ec25631f63135

I will squash once I get all the +1s.

nmattam:
I was converging all the nodes for one last time and noticed these error on the forwarders.
`11-07-2017 22:08:12.547 +0000 ERROR IndexerDiscoveryHeartbeatThread - failed to parse response payload for group=cerner_splunk/cluster-indexer-discovery, err=could not parse heartbeat period from node=18446744073709552000http_response=OK`

This is because the index_discovery stanza (`[index_discovery]`) should be set on the cluster_master even if there are no index_discovery attributes configured. While testing locally I was using the indexer discovery cluster databag on the cluster master which is why this went unnoticed. 
This is fixed here 5f6851d7026752ca771566dd2e1675ee854ff4c1. 

I created another environment file for the cluster master which uses the indexer discovery cluster databag to get the vagrant setup working.

docs/databags.md
```diff
@@ -30,6 +30,14 @@ The Cluster Hash is part of a plaintext data bag item that defines a logical gro
...
 * `['receivers']` - Array of strings of hosts where this cluster's indexers are listening. (Required for forwarders)
 * `['receiver_settings']['splunktcp']['port']` - Port indexers are listening on, and forwarders are sending data (required for forwarders and receivers)
+* `['tcpout_settings']` - Hash of tcpout settings to be configured in the outputs.conf under the `tcpout` stanza.
+* `['tcpout_settings'][???]` - Valid values are those under the tcpout stanza of [outputs.conf][]
+* `['indexer_discovery']` - Boolean to toggle indexer_discovery. Set this to `true` to leverage indexer discovery. Note that `['receivers']` configured when `['indexer_discovery']` is set to `true` will be ignored.
+* `['indexer_discovery_settings']['pass4SymmKey']` - pass4Symmkey value to be set under the `index_discovery` stanza in [server.conf][] and [outputs.conf][]
+* `['indexer_discovery_settings']['master_configs']` - Hash of indexer discovery settings to be configured on the cluster master under the `indexer_discovery` stanza.
```
gravesb: minor, would it make more sense to rename this to server_configs (since it goes in the server.conf file)?

PR #160, opened 2017-11-09
Add support for server image creation

gravesb:
In order to bake a base image for re-use in various regions in AWS, I'd like to be able to use this cookbook to lay down as much of the basic server configuration as possible.  Given that the majority of the conf files will contain some kind of environment/region specific information, this adds a new recipe with the ability to skip all the configuration other than apps.  Apps certainly may contain environment specific information, but they can also be generic skeletons that can apply to multiple environments.


gravesb:
> Rubocop is complaining and I think we also need some README updates about

Fixed foodcritic and Updated doc: https://github.com/cerner/cerner_splunk/pull/160/commits/2ca24be83acc1e59a9d8e65cdd704be5ca1af882

recipes/image_prep.rb
```diff
@@ -16,3 +16,10 @@
   notifies :stop, 'service[splunk]', :before
   notifies :delete, 'file[splunk-marker]', :before
 end
+
+execute 'clean eventdata' do
+  command "#{node['splunk']['cmd']} clean eventdata -f"
+  user node['splunk']['user']
+  group node['splunk']['group']
+  only_if { node['splunk']['node_type'] != :forwarder }
```
acharlieh: I wonder if we should be using the nodetype or if we should use the package base name... as the heavy forwarder recipe also sets node_type forwarder: https://github.com/cerner/cerner_splunk/blob/2.14.0/recipes/heavy_forwarder.rb#L9

recipes/_configure.rb
```diff
@@ -5,25 +5,28 @@
 #
 # Configures the Splunk system post package installation
 
-# Verify that clusters are configured
-if node['splunk']['node_type'] != :license_server && node['splunk']['config']['clusters'].empty?
-  if node['splunk']['node_type'] == :forwarder
-    Chef::Log.warn 'No cluster data bag configured, ensure your outputs are configured elsewhere.'
-  else
-    throw 'You need to configure at least one cluster databag.'
+unless node['splunk']['configure_apps_only']
```
gravesb: If we want this to be less reusable outside of the new server_install_only recipe and more of an internal implementation detail, I could move this to node.run_state instead of an attribute.

gravesb: Moved to run_state: https://github.com/cerner/cerner_splunk/pull/160/commits/13c0f592422824b153bd1d9caf4b72d0dbd8b825

recipes/server_install_only.rb
```diff
@@ -0,0 +1,17 @@
...
+#
+# Only run the basic install for the purposes of baking a base server image. Skips
+# the vast majority of configuration since it's likely environment specific.
+#
+
+fail 'Server installation not currently supported on windows' if platform_family?('windows')
+
+## Attributes
+instance_exec :server, &CernerSplunk::NODE_TYPE
+node.run_state['cerner_splunk'] ||= {}
```
acharlieh: Very minor, and possibly bikeshedding... but if we put this in an attributes file, could we then just use it anywhere without the extra falsy checks?

gravesb: Seems to work, and reduces a few lines of code: https://github.com/cerner/cerner_splunk/pull/160/commits/6d076e50ee34a53b13a17c3007a7e083715893d9

gravesb:
Rebased and squashed commits.

PR #163, opened 2018-01-12
Upgrade to Splunk 7.0.1

nmattam:
PR upgrades Splunk to 7.0.1. Since SHC captain now syncs the admin password with the members, this PR also configures the admin password for these members.

### DIFF
--- a/vagrant_repo/environments/splunk_server.json
+++ b/vagrant_repo/environments/splunk_server.json
@@ -16,8 +16,15 @@
           "c2_captain": "cerner_splunk/passwords:splunk.secret",
           "c2_deployer": "cerner_splunk/passwords:splunk.secret",
           "c2_newnode": "cerner_splunk/passwords:splunk.secret"
+        },
+        "admin_password": {
+          "c2_boot1": "cerner_splunk/standalone_passwords:admin_password",
+          "c2_boot2": "cerner_splunk/standalone_passwords:admin_password",
+          "c2_captain": "cerner_splunk/standalone_passwords:admin_password",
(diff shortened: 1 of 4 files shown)


vagrant_repo/environments/splunk_server.json
```diff
@@ -16,8 +16,15 @@
           "c2_captain": "cerner_splunk/passwords:splunk.secret",
           "c2_deployer": "cerner_splunk/passwords:splunk.secret",
           "c2_newnode": "cerner_splunk/passwords:splunk.secret"
+        },
+        "admin_password": {
+          "c2_boot1": "cerner_splunk/standalone_passwords:admin_password",
+          "c2_boot2": "cerner_splunk/standalone_passwords:admin_password",
+          "c2_captain": "cerner_splunk/standalone_passwords:admin_password",
+          "c2_newnode": "cerner_splunk/standalone_passwords:admin_password"
```
gravesb: You don't have to set these the same right?  Is it just that if you don't, then when you run chef-client it will update the password, only to have it changed back to what the rest of the cluster agrees on in the next sync?

nmattam: Yes and it will write a password to `/etc/splunk/password` which is not the actual `admin` password.

acharlieh: I think this is ok for now, to force the passwords to match, as when what we think the old password should be versus what the password actually is don't match that would cause havoc with the chef run. 

Changing this password I think is problematic in an environment where it's now synced between nodes.

Issue #164, opened 2018-01-16
Vagrant provision fails on the windows forwarder


nmattam:
Vagrant provision fails on the windows forwarder with the following error.

> ==> f_win2012r2: Running chef-client...
> ==> f_win2012r2: [2018-01-16T10:38:52-08:00] FATAL: Configuration error NoMethodError: undefined method `<<' for nil:NilClass
> ==> f_win2012r2: [2018-01-16T10:38:52-08:00] FATAL:   C:/vagrant-chef/client.rb:32:in `from_string'
> ==> f_win2012r2: [2018-01-16T10:38:52-08:00] FATAL: Aborting due to error in 'C:/vagrant-chef/client.rb'
> 

It was noticed that chef-client 13.6.4 was installed on this box even though the [omnibus version is set to 12](https://github.com/cerner/cerner_splunk/blob/2b1b14f9626b40931564313bdad832d64747aecd/Vagrantfile#L47). 

> C:\Users\vagrant>                                                                                                                                                                                                                                                             
> C:\Users\vagrant>chef-client -v                                                                                                                                                                                                                                               
> Chef: 13.6.4    

This issue is reported [here](https://github.com/chef/vagrant-omnibus/issues/129) and a workaround is suggested [here](https://github.com/chef/vagrant-omnibus/issues/118#issuecomment-228749135). Following the workaround, I was able to [successfully provision the windows forwarder](https://gist.github.com/nmattam/a1e175c4cc53ab65949e04232adf6a53).

Note:
I used a chef-client version <12.19 because of https://github.com/cerner/cerner_splunk/issues/142.

nmattam:
PR: https://github.com/cerner/cerner_splunk/pull/165

PR #165, opened 2018-01-17
Fix windows forwarder converge error

nmattam:
Fixes https://github.com/cerner/cerner_splunk/issues/164 for the time being.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -350,6 +350,9 @@ Vagrant.configure('2') do |config|
     # Reference: https://github.com/chef/vagrant-omnibus/issues/90#issuecomment-51816397
     cfg.vm.guest = :windows
     default_omnibus config
+    # config below prevents the installation of latest Chef on the box.
+    # Reference: https://github.com/chef/vagrant-omnibus/issues/118
+    config.omnibus.install_url = 'https://packages.chef.io/files/stable/chef/12.18.31/windows/2012r2/chef-client-12.18.31-1-x64.msi'
     cfg.vm.provider :virtualbox do |vb|
       vb.customize ['modifyvm', :id, '--memory', 1024]
     end


Vagrantfile
```diff
@@ -350,6 +350,9 @@ Vagrant.configure('2') do |config|
     # Reference: https://github.com/chef/vagrant-omnibus/issues/90#issuecomment-51816397
     cfg.vm.guest = :windows
     default_omnibus config
+    # config below prevents the installation of latest Chef on the box.
+    # Reference: https://github.com/chef/vagrant-omnibus/issues/129
+    config.omnibus.install_url = 'https://packages.chef.io/files/stable/chef/12.17.44/windows/2012r2/chef-client-12.17.44-1-x64.msi'
```
gravesb: Can we use 12.18.31?  That should be the latest working version wrt windows.

Vagrantfile
```diff
@@ -350,6 +350,9 @@ Vagrant.configure('2') do |config|
     # Reference: https://github.com/chef/vagrant-omnibus/issues/90#issuecomment-51816397
     cfg.vm.guest = :windows
     default_omnibus config
+    # config below prevents the installation of latest Chef on the box.
+    # Reference: https://github.com/chef/vagrant-omnibus/issues/129
```
gravesb: Minor, issue 129 was commented as a duplicate of issue 118.  Since most of the useful comments and workarounds are on 118, I would prefer to reference that issue instead.

nmattam: c56339bcc3223c7413cf699102abb7237d3f8049

Vagrantfile
nmattam: c56339bcc3223c7413cf699102abb7237d3f8049

PR #170, opened 2018-07-30
ETSTOOLS-9225: Changed search time to -1h

Cam-Barker:
Attribute "dispatch.earliest_time" was changed form @d to -1h to potentially speed up search times.

### DIFF
--- a/attributes/_configure.rb
+++ b/attributes/_configure.rb
@@ -4,7 +4,7 @@ default['splunk']['config']['host'] = node['ec2'] ? node['ec2']['instance_id'] :
 
 default['splunk']['config']['licenses'] = nil
 default['splunk']['config']['ui_prefs']['default'] = {
-  'dispatch.earliest_time' => '@d',
+  'dispatch.earliest_time' => '-1h',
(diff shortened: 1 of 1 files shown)


gravesb:
-1.  There are a dozen ways you can set this attribute at a higher precedence if want it to be a different value. The existing default seems perfectly reasonable.

https://docs.chef.io/attributes.html#attribute-precedence

acharlieh:
I would agree with @gravesb... When running a small system, the default of "all time" that comes with Splunk is the most reasonable as you can see all data easily... medium systems the default of Today as was here previously is more reasonable... As you wind up with growing installations, there's often a progression to smaller and smaller default search times. 

Being an open source project I feel like keeping the reasonable default of Today default search time that we've had is most appropriate, enabling overriding of this default in attributes / role cookbooks for a particular installation. 

If we were to take this, being a change to a default that could impact functionality, I feel like it would necessitate a major version bump... which could be done, but it feels silly lacking a number of other major changes going at the same time (Chef 14+ support for example)

PR #173, opened 2018-09-19
Adding support for fixed-sourcetype licenses 

paul-hoff:
Adding some logic to support a fixed-sourcetype license


recipes/license_server.rb
```diff
@@ -31,7 +31,7 @@
     type = doc.at_xpath('/license/payload/type/text()').to_s
     quota = doc.at_xpath('/license/payload/quota/text()').to_s.to_i
     expiration_time = doc.at_xpath('/license/payload/expiration_time/text()').to_s.to_i
-    total_available_license_quota += quota if type == 'enterprise' && expiration_time > Time.now.to_i
+    total_available_license_quota += quota if type == 'enterprise' || type == 'fixed-sourcetype' && expiration_time > Time.now.to_i
```
gravesb: for the sake of clarity/readability, put parens around the two type == checks.
`if (type == 'enterprise' || type == 'fixed-sourcetype') && expiration_time > Time.now.to_i`

recipes/license_server.rb
```diff
@@ -31,7 +31,7 @@
     type = doc.at_xpath('/license/payload/type/text()').to_s
     quota = doc.at_xpath('/license/payload/quota/text()').to_s.to_i
     expiration_time = doc.at_xpath('/license/payload/expiration_time/text()').to_s.to_i
-    total_available_license_quota += quota if type == 'enterprise' || type == 'fixed-sourcetype' && expiration_time > Time.now.to_i
+    total_available_license_quota += quota if type == ('enterprise' || type == 'fixed-sourcetype') && expiration_time > Time.now.to_i
```
gravesb: parens are in the wrong location.

acharlieh:
This isn't the only change to support fixed-sourcetype licenses... the licenses are put into a different directory, and the stack_ids are generated. For the single sourcetype case, it seems to be fixed-sourcetype_UPPERCASE(SHA256(sourcetype)) but a question is what happens with multiple.

recipes/license_server.rb
```diff
@@ -29,22 +30,31 @@
   unless %w[id chef_type data_bag].include? key
     doc = Nokogiri::XML value
     type = doc.at_xpath('/license/payload/type/text()').to_s
+    sourcetype_name = doc.search('//sourcetypes/*').map(&:text) if type == 'fixed-sourcetype'
```
acharlieh: how about: 

```ruby
sourcetypes = doc.search('//sourcetype').map(&:text).join
type = "#{type}_#{Digest::SHA256.hexdigest(sourcetypes).upcase}" if type == 'fixed-sourcetype'
```

recipes/license_server.rb
```diff
@@ -65,8 +75,8 @@
 
 b = ruby_block 'license cleanup' do
```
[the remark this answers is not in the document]
paul-hoff: I changed it back to what it originally. I might be missing something, but wont that already cleanup the license groups that are no longer used?

recipes/license_server.rb
```diff
@@ -25,13 +26,15 @@
...
   unless %w[id chef_type data_bag].include? key
     doc = Nokogiri::XML value
-    type = doc.at_xpath('/license/payload/type/text()').to_s
+    nodde.run_state['type'] = doc.at_xpath('/license/payload/type/text()').to_s
+    sourcetypes = doc.search('//sourcetype').map(&:text).join
+    node.run_state['type'] = "#{node.run_state['type']}_#{Digest::SHA256.hexdigest(sourcetypes).upcase}" if node.run_state['type'] == 'fixed-sourcetype'
     quota = doc.at_xpath('/license/payload/quota/text()').to_s.to_i
     expiration_time = doc.at_xpath('/license/payload/expiration_time/text()').to_s.to_i
-    total_available_license_quota += quota if type == 'enterprise' && expiration_time > Time.now.to_i
+    total_available_license_quota += quota if (type == 'enterprise' || type == 'fixed-sourcetype') && expiration_time > Time.now.to_i
```
nmattam: We don't have a `type` variable anymore.

paul-hoff:
I verified that I can successfully converge the license server 

`==> s_license: Chef Client finished, 36/52 resources updated in 11 minutes 58 seconds`

recipes/_configure_server.rb
```diff
@@ -249,11 +249,11 @@
 
   license_pools['pools'].each do |pool, pool_config|
     pool_max_size = CernerSplunk.convert_to_bytes pool_config['size']
-    server_stanzas["lmpool:#{pool}"] = {
+    server_stanzas["lmpool:#{node.run_state['type']}"] = {
```
nmattam: Won't this value be empty because we are evaluating this at the compile time? Can you show what the server.conf looks like on the license server?

Gemfile
```diff
@@ -14,5 +14,6 @@ gem 'chef', chef_version
 gem 'chef-sugar'
 gem 'chef-vault', chef_vault_version
 gem 'chefspec', chefspec_version
+gem 'cucumber-core', '~> 3.2'
```
nmattam: Add https://github.com/cucumber/cucumber-ruby-core/issues/160 as a comment so that we know why we are locking this.

.rubocop.yml
```diff
@@ -33,3 +33,6 @@ Style/TernaryParentheses:
 
 Style/VariableNumber:
   Enabled: false
+
+Lint/UselessAssignment:
```
nmattam: Why was this added?

paul-hoff: I added this for a test and forgot to remove it. I will remove it.

recipes/_configure_server.rb
nmattam: Also, are we grouping pools by license type?

paul-hoff: Talking with Charlie, he mentioned that for the fixed-sourcetype license the folder and pool are the same, which is fixed-sourcetype_Uppercase(sha256(sourcetypeName)). So, I was trying to account for that here

recipes/license_server.rb
```diff
@@ -25,15 +26,20 @@
...
+    node.run_state['type'] = doc.at_xpath('/license/payload/type/text()').to_s
+    node.run_state['type'] = "#{node.run_state['type']}_#{Digest::SHA256.hexdigest(sourcetypes).upcase}" if node.run_state['type'] == 'fixed-sourcetype'
     quota = doc.at_xpath('/license/payload/quota/text()').to_s.to_i
     expiration_time = doc.at_xpath('/license/payload/expiration_time/text()').to_s.to_i
-    total_available_license_quota += quota if type == 'enterprise' && expiration_time > Time.now.to_i
-    hash[type] ||= {}
-    hash[type][key] = value
+    total_available_license_quota += quota if (node.run_state['type'] == 'enterprise' || node.run_state['type'] == "fixed-sourcetype_#{Digest::SHA256.hexdigest(sourcetypes).upcase}") && expiration_time > Time.now.to_i
+    unless hash[node.run_state['type']].nil?
+      fail 'Multiple license types are not currently supported' if hash[node.run_state['type']] != value
```
nmattam: Not sure I follow why we are comparing type to license value, which is the actual license. I think we should be comparing each license type with the license types already in the hash (unless the hash is empty) and fail if we find a different license type.

nmattam: Also, instead of updating node.run_state every time, how about we just read the value of type into a variable and use it inside the loop. After the loop, we can assign this variable to node.run_state['type'].

paul-hoff: Should that looking something like this?
`    unless hash[type].nil?
      fail 'Multiple license types are not currently supported' if hash[type] != type
    end`

paul-hoff: I converged the node successfully and verified that the server.conf and licenses were created as expected.

recipes/_configure_server.rb
```diff
@@ -246,16 +246,19 @@
...
+    block do
+      license_pools['pools'].each do |pool, pool_config|
+        pool_max_size = CernerSplunk.convert_to_bytes pool_config['size']
+        server_stanzas["lmpool:#{pool}"] = {
+          'description' => pool,
+          'quota' => pool_max_size,
+          'slaves' => pool_config['GUIDs'].join(','),
+          'stack_id' => node.run_state['type']
+        }
+        allotted_pool_size += pool_max_size
```
gravesb: Since this moves the allotted_pool_size summation from compile time to execution time, I believe that makes the run state for the total_allotted_pool_size incorrect (since the node.run_state is set below, outside the ruby_block at compile time).  Then the error checking on line 48-51 of the license_server recipe is no longer accurate.

If one part of it moves to execution time, then all the parts have to move.  But then it won't fail the chef run until part way through, which makes that error checking considerably less useful since the bad server conf would have already been written out.  Will Splunk fail to start if the pools configured in server.conf are greater than the available licenses?

gravesb: I checked and Splunk doesn't seem to care if the total allocation of the pools is greater than the amount of the license stack. Although if you create or edit a pool via the UI and try to make it larger, something in the web console will set the pool size equal to the amount of available license when it saves it.

gravesb: I think the easiest fix here is to remove the ruby block here and simply re-order the operations in the license_server recipe so the data bag is processed prior to including this recipe. Just move line 18 down between lines 46 and 48 in the license_server recipe.

recipes/license_server.rb
```diff
@@ -25,19 +23,28 @@
...
     doc = Nokogiri::XML value
+    sourcetypes = doc.search('//sourcetype').map(&:text).join
     type = doc.at_xpath('/license/payload/type/text()').to_s
+    type = "#{type}_#{Digest::SHA256.hexdigest(sourcetypes).upcase}" if type == 'fixed-sourcetype'
     quota = doc.at_xpath('/license/payload/quota/text()').to_s.to_i
     expiration_time = doc.at_xpath('/license/payload/expiration_time/text()').to_s.to_i
-    total_available_license_quota += quota if type == 'enterprise' && expiration_time > Time.now.to_i
+    total_available_license_quota += quota if (type == 'enterprise' || type == "fixed-sourcetype_#{Digest::SHA256.hexdigest(sourcetypes).upcase}") && expiration_time > Time.now.to_i
     hash[type] ||= {}
+    unless hash[type].empty?
```
gravesb: this isn't currently working as intended.  This should work:
```
    if hash.length > 1
      fail 'Multiple license types are not currently supported'
    end
```

gravesb: no, the hash[type] ||= {} is sufficient to add an entry to the hash.

paul-hoff:
Should I use some sort of regular expression on line 76 of license-server.rb instead of the type variable to fix lines 73-87?

recipes/license_server.rb
nmattam: That is true. Didn't see that.

gravesb:
I think we could add something like this to that ruby block:
```ruby
Dir.glob(""#{node['splunk']['home']}/etc/licenses/*") { |dir| FileUtils.rm_rf(dir) unless dir.end_with?(type) }
```
may also need to require 'fileutils' somewhere in there along with similar logging to what's there for the other files we delete.

paul-hoff:
I have added this code to the ruby block, and was able to successfully converge the license server. I am unsure how to test its actual functionality since only a single demo license is being used.  Is this kind of what you had in mind?

```      
      existing_directory = Dir.glob("#{node['splunk']['home']}/etc/licenses/*")
      delete_directory = existing_directory.delete_if { |dir| FileUtils.rm_rf(dir) unless dir.end_with?(type) }
      delete_directory.each do |directory|
        Chef::Log.info("ruby_block[license cleanup] deleted unconfigured license directory #{directory}")
      end
```

gravesb:
yeah, something like that.  Given that you're working on a PR for fixed-sourcetype license support I assume you have access to other licenses that you could use to test locally (though obviously you wouldn't want to commit those changes as part of this PR).

recipes/license_server.rb
```diff
@@ -66,6 +72,11 @@
 b = ruby_block 'license cleanup' do
   block do
     license_groups.each do |type, licenses|
+      existing_directory = Dir.glob("#{node['splunk']['home']}/etc/licenses/*")
```
acharlieh: I know that we only support one type now, but this logically should be outside of the license_groups.each loop

recipes/license_server.rb
```diff
@@ -65,7 +71,12 @@
 
 b = ruby_block 'license cleanup' do
   block do
+    existing_directory = Dir.glob("#{node['splunk']['home']}/etc/licenses/*")
     license_groups.each do |type, licenses|
+      existing_directory.each do |dir|
```
acharlieh: Pull this loop out of the license_groups loop too...

recipes/license_server.rb
```diff
@@ -65,6 +71,11 @@
 
 b = ruby_block 'license cleanup' do
   block do
+    existing_directory = Dir.glob("#{node['splunk']['home']}/etc/licenses/*")
+    existing_directory.each do |dir|
+      FileUtils.rm_rf(dir) unless dir.end_with?(type)
```
acharlieh: Alrighty now just need to fix the logic a bit... as we want to only delete those directories that we expect there to be no licenses, so something like: (double check me on this... this is rather haphazard, but should give you the general idea)

```ruby
next if license_groups.keys.include? File.basename(dir)
Chef::Log.info( "... blah..." )
FileUtils.rm_rf(dir)
```

paul-hoff: This should be fixed with d4c6810

recipes/license_server.rb
```diff
@@ -25,19 +24,26 @@
...
+    type = "#{type}_#{Digest::SHA256.hexdigest(sourcetypes).upcase}" if type == 'fixed-sourcetype'
     quota = doc.at_xpath('/license/payload/quota/text()').to_s.to_i
     expiration_time = doc.at_xpath('/license/payload/expiration_time/text()').to_s.to_i
-    total_available_license_quota += quota if type == 'enterprise' && expiration_time > Time.now.to_i
+    total_available_license_quota += quota if (type == 'enterprise' || type == "fixed-sourcetype_#{Digest::SHA256.hexdigest(sourcetypes).upcase}") && expiration_time > Time.now.to_i
     hash[type] ||= {}
+    fail 'Multiple license types are not currently supported' if hash.length > 1
     hash[type][key] = value
   end
+  node.run_state['type'] = type
```
acharlieh: This should either move as is to right after the fail ... if on line 37 
OR
This should be outside the inject loop after the end on line 42, slightly modified like this: 
```ruby
node.run_state['type'] = license_groups.keys.first
```
That type variable defined on line 31 feels like it's really intended to only be valid within the unless block (line 28 -> 39), 

If you happen to be retrieving a level where skipping the keys 'id', 'chef_type' and 'data_bag' mattered, and one of those happened to be traversed prior to your actual key that you're interested in, you'll have a ```NameError: undefined local variable or method `type'``` on your hands

acharlieh: nit: should 'type' be 'license_type' or 'stack_id' ... something a bit more descriptive as a key?

paul-hoff: that makes sense, these changes are reflected in faaab86

gravesb:
Closed by https://github.com/cerner/cerner_splunk/commit/b426f479601064650c933cceb83b72f1b180ee74.

PR #179, opened 2019-02-13
uplifting splunk to 7.2.4 version 

srikanthreddy5:
As consumer of  cerner_splunk cookbook , uplifting  the cookbook to support  Splunk  7.2.4.


gravesb:
When Splunk starts and the user-seed.conf file exists, that file is deleted. With the current setup, even if there are no config changes you would end up causing splunk to restart unnecessarily.  We'll need a bit of additional logic to keep from writing out the user-seed if the admin user has already been initialized.  We only want to do this on the initial install.

recipes/_configure_user_seed.rb
```diff
@@ -0,0 +1,17 @@
...
+# Cookbook Name:: cerner_splunk
+# Recipe:: _configure_user_seed.rb
+#
+# Configures the system inputs.conf file
+
+# Translate monitor attributes to generic hash
+input_stanzas = {}
+input_stanzas['user_info'] = {}
+input_stanzas['user_info']['USERNAME'] = 'admin'
+input_stanzas['user_info']['PASSWORD'] = 'changeme'
```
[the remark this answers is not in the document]
gravesb: The fact that the user-seed.conf file is deleted when splunk starts makes this a bit trickier... I'm open to other opinions on this. The current logic works, it's just a little roundabout on how it gets there.

nmattam:
Came across this doc[1]. There is a `--seed-passwd` option that you can use to set the admin password. I tried passing in this option to the splunk first run command[2] and looks like it works.

[1] https://docs.splunk.com/Documentation/Splunk/7.2.4/Security/Secureyouradminaccount#Create_admin_credentials_using_the_--seed-passwd_or_--gen-and-print-passwd_CLI_arguments
[2] https://github.com/cerner/cerner_splunk/blob/483ff080983f02be1c10eea924b125000378ddb2/recipes/_install.rb#L87

srikanthreddy5:
Yes , I can do --seed-passwd in  "splunk-first-run"   
https://github.com/cerner/cerner_splunk/blob/483ff080983f02be1c10eea924b125000378ddb2/recipes/_install.rb#L87
In that  case we dont need to create  user-seed.conf file.

recipes/_install.rb
```diff
@@ -84,7 +84,7 @@ def nsp
 windows_password = CernerSplunk::DataBag.load(node['splunk']['windows_password'], secret: node['splunk']['data_bag_secret'])
 
 execute 'splunk-first-run' do
-  command "#{node['splunk']['cmd']} help commands --accept-license --answer-yes --no-prompt"
+  command "#{node['splunk']['cmd']}  start  --accept-license --answer-yes --no-prompt --seed-passwd '#{node['splunk']['package']['seed']['passwd']}' "
```
gravesb: What happens if you run this on a pre-7.2.x version of Splunk?  We should maintain backward compatibility to at least 7.0.  If the --seed-passwd is just ignored then great, otherwise we may need to flex the command based on what version of Splunk is being installed.

srikanthreddy5: updated the _install.rb file to support pre-7.2 and post 7.2

metadata.rb
```diff
@@ -4,7 +4,7 @@
 license          'Apache-2.0'
 description      'Installs/Configures Splunk Servers and Forwarders'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version          '2.24.0'
+version          '2.24.1'
```
gravesb: I think 2.25.0 would be more appropriate.

recipes/_install.rb
```diff
@@ -83,10 +83,15 @@ def nsp
 
 windows_password = CernerSplunk::DataBag.load(node['splunk']['windows_password'], secret: node['splunk']['data_bag_secret'])
 
+run_command = "#{node['splunk']['cmd']} help commands --accept-license --answer-yes --no-prompt"
+if Gem::Version.new(nsp['version']) >= Gem::Version.new('7.2.0')
+  run_command += " --seed-passwd 'changeme'"
```
gravesb: Ideally, I still think this should be the configured admin password (if there is one) rather than hard coding to 'changeme' and then letting the _generate_password recipe update it later on in the chef run. Functionally it all ends up the same though.  @acharlieh @nmattam thoughts?

acharlieh: So I kinda agree with @gravesb on this, as some of the point of this feature I feel was to eliminate the changeme password even from the the very first start of the software and doing this feels like it kinda bypasses the control that was intended.

That said, arguably this cookbook for years has helped ensure that there was no default password being used, and (without reviewing the rest of this proposed change), it seems that it will continue to do so. So it might be an acceptable transitionary point for the time being, until say there could be a feature to work towards eliminating the `/etc/splunk/password` file :) 

I will say though, if you do keep the route of hardcoding the seed password to changeme, I would argue this is not a sensitive resource. (changeme is a well known default, and this is an open source project...) and should something go sideways with this command, non-sensitive resources are easier to debug.

acharlieh:
Two other operational pains you may want to think about 1) resetting out of sync admin passwords is a bit more difficult in 7.1+ this blog might give some ideas: https://www.hurricanelabs.com/splunk-tutorials/splunk-7-1-performing-a-splunk-password-reset

and 2) Did you see that secrets in 7.2 conf files are $7$ now? I'm just going to leave this here: https://github.com/HurricaneLabs/splunksecrets/blob/master/splunksecrets.py#L40

recipes/_install.rb
srikanthreddy5: I agree with your suggestions but if we plan to use the admin_password directly for splunk-first-run for Splunk 7.2, We have to add a condition not to execute the _generate_password.rb for Splunk 7.2 and use almost all code from the _genereate_password.rb  except change password for Splunk 7.2.x  like
    1) creating  /etc/splunk/password
    2) adding new_password to file.
Instead of duplication, until all our environment upgraded to 7.2, we can use this way. Once our environment is upgraded we come up with an efficient approach.

gravesb: I'm ok with removing the sensitive true flag then and we'll plan to circle back later and rework the _generate_password recipe. Might be good to throw a TODO comment in here as a reminder.

srikanthreddy5: Removed sensitive true flag from splunk-first-run (_install.rb) and added TODO comment.

recipes/_install.rb
```diff
@@ -83,8 +83,13 @@ def nsp
...
+end
+#TODO: Use admin_password from databag for splunk-first-run.
```
gravesb: minor, need a space between `#` and `TODO`.  This is currently failing the build due to the rubocop offense.

srikanthreddy5: added space between # and TODO .Thanks

gravesb:
@nmattam I'd like to get one more approval before I merge this. Can you take another look?

gravesb:
Yeah, I tested as well and the encryption seems to maintain backward compatibility for now. If they're encrypted with the old style Splunk can still read and decrypt them fine.  If you change it to plaintext and restart Splunk, it will encrypt with the new algorithm though.

PR #182, opened 2019-03-11
Updated Splunk patch version 7.2.4.2 and added attribute for smartstore 

srikanthreddy5:
Updating the Splunk 7.2.4.2 and adding an attribute to enable Smartstore(S2).

Added a condition for "tstatsHomePath", as it has to remain unset to enable S2. 
https://docs.splunk.com/Documentation/Splunk/7.2.4/Indexer/ConfigureSmartStore

### DIFF
--- a/docs/databags.md
+++ b/docs/databags.md
@@ -80,11 +80,11 @@ An Indexes Hash is part of a plaintext data bag item that defines the set of ind
     * `_directory_name` - The on-disk name of the directory to store the index. Defaults to the index name.
     * `_maxDailyDataSizeMB` - The amount of daily usage this index is expected to consume.  Used to calculate the maxTotalDataSizeMB if maxTotalDataSizeMB has not already been specified for the index.
     * `_dataSizePaddingPercent` - The percentage of padding to apply to the amount of space this index is expected to consume.  Used to calculate the maxTotalDataSizeMB if maxTotalDataSizeMB has not already been specified for the index. Defaults to 10 if no value is specified.
+    * `_noGenerateTstatsHomePath` - Do not generate tstatsHomePath for this index when set to true. Used to disable datamodel acceleration in case of the smartstore enabled indexes. 
 * `['flags']` - These define boolean processing flags per index. All flags are default 'false' but can be set to true. Current flags include:
     * `noGeneratePaths` - Do not generate the homePath,coldPath,thawedPath to this index when not present in the config above
     * `noRepFactor` - Do not add 'repFactor = auto' to this index when not present in the config on a cluster master.
 * `['metadata']` - These define ownership / other reference metadata around indexes and their owners (ALPHA!!! CAN CHANGE!!!)
-* '_noGenerateTstatsHomePath' - for smartstore enabled indexes tstatsHomePath should be unset.If _noGenerateTstatsHomePath is true in default stanza it means smartstore enabled on all the indexes and on individual indexes smart store can be enabled for per Index. Default is false/nil.
 
 Roles Hash
 ----------
--- a/docs/databags.md
+++ b/docs/databags.md
@@ -84,7 +84,7 @@ An Indexes Hash is part of a plaintext data bag item that defines the set of ind
     * `noGeneratePaths` - Do not generate the homePath,coldPath,thawedPath to this index when not present in the config above
     * `noRepFactor` - Do not add 'repFactor = auto' to this index when not present in the config on a cluster master.
 * `['metadata']` - These define ownership / other reference metadata around indexes and their owners (ALPHA!!! CAN CHANGE!!!)
-
+* '_noGenerateTstatsHomePath' - for smartstore enabled indexes tstatsHomePath should be unset.If _noGenerateTstatsHomePath is true in default stanza it means smartstore enabled on all the indexes and on individual indexes smart store can be enabled for per Index. Default is false/nil.
 Roles Hash
 ----------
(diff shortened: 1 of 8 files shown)


recipes/_configure_indexes.rb
```diff
@@ -63,7 +63,7 @@
       hash['coldPath'] = "#{base_path}/#{dir_name}/colddb" unless hash['coldPath']
       hash['homePath'] = "#{base_path}/#{dir_name}/db" unless hash['homePath']
       hash['thawedPath'] = "$SPLUNK_DB/#{dir_name}/thaweddb" unless hash['thawedPath']
-      hash['tstatsHomePath'] = "#{base_path}/#{dir_name}/datamodel_summary" if volume && !hash['tstatsHomePath']
+      hash['tstatsHomePath'] = "#{base_path}/#{dir_name}/datamodel_summary" if volume && !hash['tstatsHomePath'] && node['splunk']['enable_smart_store'] == false
```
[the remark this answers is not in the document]
srikanthreddy5: updated the cookbook to support smart-store per index.

docs/databags.md
```diff
@@ -84,6 +84,7 @@ An Indexes Hash is part of a plaintext data bag item that defines the set of ind
     * `noGeneratePaths` - Do not generate the homePath,coldPath,thawedPath to this index when not present in the config above
     * `noRepFactor` - Do not add 'repFactor = auto' to this index when not present in the config on a cluster master.
 * `['metadata']` - These define ownership / other reference metadata around indexes and their owners (ALPHA!!! CAN CHANGE!!!)
+* '_noGenerateTstatsHomePath' - for smartstore enabled indexes tstatsHomePath should be unset.If _noGenerateTstatsHomePath is true in default stanza it means smartstore enabled on all the indexes and on individual indexes smart store can be enabled for per Index. Default is false/nil.
```
nmattam: This should go under `configs`. Also, the description kind of sends the message that setting this attribute will enable smart store. This attribute will only let you not set the tstatshomepath and enabling smart store will require configuring other attributes. How about something like
> Do not generate tstatsHomePath for this index when set to true.

srikanthreddy5: updated the comment and moved under config. Thanks

metadata.rb
```diff
@@ -4,7 +4,7 @@
 license          'Apache-2.0'
 description      'Installs/Configures Splunk Servers and Forwarders'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version          '2.25.0'
+version          '2.25.1'
```
gravesb: This should be 2.26.0 since the change is an enhancement rather than a bug fix.

recipes/_configure_indexes.rb
```diff
@@ -63,7 +63,7 @@
       hash['coldPath'] = "#{base_path}/#{dir_name}/colddb" unless hash['coldPath']
       hash['homePath'] = "#{base_path}/#{dir_name}/db" unless hash['homePath']
       hash['thawedPath'] = "$SPLUNK_DB/#{dir_name}/thaweddb" unless hash['thawedPath']
-      hash['tstatsHomePath'] = "#{base_path}/#{dir_name}/datamodel_summary" if volume && !hash['tstatsHomePath']
+      hash['tstatsHomePath'] = "#{base_path}/#{dir_name}/datamodel_summary" if volume && !hash['tstatsHomePath'] && !no_gentstat
```
gravesb: We should unit test(s) to cover this new functionality. Add to the _configure_indexes_spec.

metadata.rb
srikanthreddy5: updated the version

recipes/_configure_indexes.rb
srikanthreddy5: I have added new unit tests.

spec/unit/recipes/_configure_indexes_spec.rb
```diff
@@ -134,6 +134,91 @@
         }
       }
 
+      expect(subject).to create_splunk_template('system/indexes.conf').with(expected_attributes)
+    end
+  end
+  # When _noGenerateTstatsHomePath => true in specific indexes, then writes the indexes.conf file without tstats for only for that indexes
```
nmattam: Instead of writing this as a comment, can you update the `context` and `it` accordingly.

```
context 'when _noGenerateTstatsHomePath is set to true for specific indexes' do
    it 'writes the indexes.conf file without tstatsHomePath only for those indexes' do
```

Same comment applies to the next test.

spec/unit/recipes/_configure_indexes_spec.rb
```diff
@@ -134,6 +134,91 @@
...
+  context 'when _noGenerateTstatsHomePath is set to true' do
+    let(:index_config) do
+      {
+        'config' => {
+          'volume:test' => {
+            'path' => '/test/path'
+          },
+          'index_a' => { '_directory_name' => 'foo', '_noGenerateTstatsHomePath' => true },
```
nmattam: if `_volume` is not set, then `tstatsHomePath` will not get set[1]. So you need to add a `_volume` attribute to really test `_noGenerateTstatsHomePath`. Same applies to the other test.
[1] https://github.com/cerner/cerner_splunk/pull/182/files#diff-79be619f4d42bccfc1205531edfcace2R66

spec/unit/recipes/_configure_indexes_spec.rb
srikanthreddy5: removed the comment and updated in context and it.

spec/unit/recipes/_configure_indexes_spec.rb
srikanthreddy5: updated _volume. Thanks

PR #184, opened 2019-04-09
Update to work on el 7.x with Splunk 7.2.2+

### LINKED ISSUE #181: Splunk 7.2 support doesn't include systemd changes
7.2 support was added in #179, however starting in 7.2.2 Splunk started generating a systemd unit file with the enable-boot start command ([link](https://docs.splunk.com/Documentation/Splunk/7.2.2/Admin/RunSplunkassystemdservice)).

```
[root@rhel7test ~]# rpm -q splunkforwarder
splunkforwarder-7.1.6-8f009a3f5353.x86_64
[root@rhel7test ~]# /opt/splunkforwarder/bin/splunk enable boot-start
Init script installed at /etc/init.d/splunk.
Init script is configured to run at boot.
```
vs.
```
[root@rhel7test ~]# rpm -q splunkforwarder
splunkforwarder-7.2.4-8a94541dcfac.x86_64
[root@rhel7test ~]# /opt/splunkforwarder/bin/splunk enable boot-start
Init script installed at /etc/systemd/system/.
Init script is configured to run at boot.
```

This causes the [_start](https://github.com/cerner/cerner_splunk/blob/2aa5ac6390a2cf3d304bdc2ad34fc246999d4cde/recipes/_start.rb) recipe to fail since it's hard-coded to look for /etc/init.d/splunk. This only affects net-new installs since upgrades would leave the old init script behind. That does bring up another question though, should the cookbook handle removing the old init script once the systemd one is landed?

gravesb:
Fixes #181 

In the current landscape, there are a number of changes that would need to be made to work under systemd.  As an alternative, we can simply tell splunk not to use systemd and we can continue running under init.d for the foreseeable future.  I expect at some point down the road we may want to do a larger refactor to allow for running under sytemd, but for now this will work.

On an unrelated note, I also bumped the splunk version to 7.2.5.


attributes/_install.rb
```diff
@@ -13,8 +13,8 @@
     '/etc/splunk'
   end
 
-default['splunk']['package']['version'] = '7.2.4.2'
-default['splunk']['package']['build'] = 'fb30470262e3'
+default['splunk']['package']['version'] = '7.2.5'
```
nmattam: There appears to be a bug in 7.2.5 with the FIELDALIAS in props.conf, causing search heads to crash[1]. Should we go for the patch version? [2].

[1] https://answers.splunk.com/answers/737075/search-head-crashing-after-upgrade-to-725-due-to-f.html
[2] https://docs.splunk.com/Documentation/Splunk/7.2.5/ReleaseNotes/Fixedissues#Splunk_Enterprise_7.2.5.1

gravesb: Updated to 7.2.5.1 in the latest commit.

Vagrantfile
```diff
@@ -301,6 +301,7 @@ Vagrant.configure('2') do |config|
 
   config.vm.define :s_standalone do |cfg|
     default_omnibus config
+    config.vm.box = 'bento/centos-7.6'
```
gravesb: Thoughts on just changing line 78 to centos-7.6 for all the boxes instead of just overriding it for these two?

spec/cookbooks/cerner_splunk_test/recipes/_splunk_service.rb
```diff
@@ -1,22 +0,0 @@
-# coding: UTF-8
```
gravesb: deleted this file since it didn't appear to be used anywhere.

nmattam:
@gravesb I'm seeing some random errors trying to converge the forwarder. Different errors I'm seeing: https://gist.github.com/nmattam/8ec723bb637f98391a69ae6ea4caf274. Although, I don't see these on the 7.6 standalone. Also, I'm not seeing this on the 6.7 forwarder. Looks like the chef process is getting killed in the middle of the run. Did you run into this?

Vagrantfile
nmattam: +1

gravesb:
Yeah, I ran into similar issues with only the forwarder.  Removing the cerner_splunk_test recipe for the initial converge seems to allow it to complete more often. But then adding it back in and running provision, the chef run still completes so that recipe alone doesn't seem to be the problem.  My first several attempts all seemed to fail, but the last ones I ran succeeded more often so it's a very inconsistent problem. I was hoping it was an issue with my network or vagrant setup, but that doesn't appear to be the case.  It doesn't even die at the same point every time, I'm not really sure what's going on.

nmattam:
Saw this on the forwarder box https://gist.github.com/nmattam/8f0d42f0e75ef6817beed73bba3fa05d.
Looks like we are running out of memory. We have 128MB set on the forwarders and 256MB on the standalone, which explains why this happens only on the forwarder. I verified that I can successfully converge the forwarder with 256MB. Do you want to bump up the memory?

Vagrantfile
gravesb: Done in the latest commit.

gravesb:
Nice find.  Weird that the s_license would come up without problems on 128Mb memory but the forwarder wouldn't.  I went ahead and just bumped them all up to 256 in the latest commit since there were several still using 128.

Issue #107, opened 2016-03-14
Lower/configure internal log rotation limits


mattnelson:
The internal logs of the splunk forwarder have the potential to take up to a large amount of the filesystem. The default maxFileSize and maxBackupIndex should either be lowered or support configuration to a more sensible value.

```
14:56:43 $ grep -E "maxFileSize|maxBackupIndex" /opt/splunkforwarder/etc/log.cfg
# if these log files are getting too big for your liking, turn down the maxFileSize.
appender.A1.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.A1.maxBackupIndex=5
appender.license_usage.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.license_usage.maxBackupIndex=5
appender.A4.maxFileSize=0
appender.metrics.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.metrics.maxBackupIndex=5
appender.audittrail.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.audittrail.maxBackupIndex=5
appender.licenseaudit.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.licenseaudit.maxBackupIndex=5
appender.accesslog.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.accesslog.maxBackupIndex=5
appender.uiaccess.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.uiaccess.maxBackupIndex=5
appender.scheduler.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.scheduler.maxBackupIndex=5
appender.remotesearches.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.remotesearches.maxBackupIndex=5
appender.idata_ResourceUsage.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.idata_ResourceUsage.maxBackupIndex=5
appender.conf.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.conf.maxBackupIndex=5
appender.idata_DiskObjects.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.idata_DiskObjects.maxBackupIndex=5
appender.idata_KVStore.maxFileSize=25000000 # default: 25MB (specified in bytes).
appender.idata_KVStore.maxBackupIndex=5
appender.kvstore_appender.maxFileSize=25000000 # default: 25MB (specified in bytes).
[...]

PR #191, opened 2019-07-12
Add attribute for splunk internal log config

rob-sch1:
We've been running into issues with log files becoming too large, as identified in this issue from a few years ago: https://github.com/cerner/cerner_splunk/issues/107. So I've added a way to configure the size and number of backups to keep of splunk's internal logs, by having the cookbook create a log-local.cfg file that overrides splunk's base log.cfg file settings (described here: https://docs.splunk.com/Documentation/Splunk/7.3.0/Troubleshooting/Enabledebuglogging#log-local.cfg).


libraries/splunk_logs.rb
```diff
@@ -0,0 +1,122 @@
...
+        file.run_action(:delete)
+        new_resource.updated_by_last_action(file.updated_by_last_action?)
+      end
+
+      def delete_file(file_path)
+        download = Chef::Resource::File.new(file_path, run_context)
+        download.run_action(:delete)
+      end
+
+      def symbolize_keys(hash)
```
rob-sch1: The manage_file/insert_procs/hash_to_proc/symbolize_keys functions are copied from libraries/splunk_apps.rb. Would it make sense to create a shared utility to host these functions, and if so what would be the right place to put it?

recipes/_configure_logs.rb
```diff
@@ -0,0 +1,13 @@
+# coding: UTF-8
+
+# Cookbook Name:: cerner_splunk
+# Recipe:: _configure_logs
+#
+# Configures the system log-local.cfg file
+
+log_local_contents = node['splunk']['logs']
+
+splunk_logs '/opt/splunk/etc/log-local.cfg' do
```
rob-sch1: I've hard-coded the location of log-local.cfg both here and in the libraries/splunk_logs.rb file. Is that acceptable or is there a way that we would prefer to define this as a relative path?

vagrant_repo/environments/splunk_standalone.json
```diff
@@ -6,6 +6,10 @@
       "apps": {
         "bag": "cerner_splunk/cluster-apps-vagrant:search_head"
       },
+      "logs": {
+        "appender.metrics.maxFileSize": "10000000"
```
kleaver: Need a comma at the end of this:
```suggestion
        "appender.metrics.maxFileSize": "10000000",
```

rob-sch1: Ah, thanks https://github.com/cerner/cerner_splunk/pull/191/commits/2ba6438e667355257343e1215d4896b0fa5b85f4

recipes/_configure_logs.rb
gravesb: The base path is configurable via `node['splunk']['home']`, prepend that to the relative location for /etc/log-local.cfg.

rob-sch1: Got it, updated here: https://github.com/cerner/cerner_splunk/pull/191/commits/0d57a4e976bd94177822cafac721537ce55cfa1c

libraries/splunk_logs.rb
```diff
@@ -0,0 +1,114 @@
...
+      def whyrun_supported?
+        true
+      end
+
+      def load_current_resource
+        @current_resource ||= Chef::Resource::SplunkLogs.new(new_resource.name)
+      end
+
+      def action_create
+        manage_file(new_resource.location, 'splunkd' => new_resource.contents)
```
gravesb: There are two stanzas in the log.cfg file, splunkd and python. Why not allow configuration of the python logging too?

libraries/splunk_logs.rb
gravesb: Do we really need this separate resource file or can we reuse the splunk_template resource in the recipe instead? e.g. 
```
splunk_template 'etc/log-local.cfg' do
  stanzas log_local_contents
  notifies :touch, 'file[splunk-marker]', :immediately
end
```

rob-sch1: As I was experimenting to learn the structure of the repo, the approaches I was trying didn't fit the mold of the template, and it didn't occur to me that they might now.

So would your suggestion be to add log-local.cfg to the list of KNOWN_CONFIG_FILES and add a regex for it to the case-when block in the template's initialize method?

gravesb: Yes, that would be my suggestion.

rob-sch1: I've done this here: https://github.com/cerner/cerner_splunk/pull/191/commits/48df09d9b795bd2b1ec22fc86b6345a788aaed42

recipes/_configure.rb
```diff
@@ -22,6 +22,7 @@
   end
 
   include_recipe 'cerner_splunk::_configure_server'
+  include_recipe 'cerner_splunk::_configure_logs'
```
nmattam: I don't think we want log-local.cfg to be created if the `node['splunk']['logs']` hash is empty or is not set. So maybe something like 
```
include_recipe 'cerner_splunk::_configure_logs' unless node['splunk']['logs'].empty?
```
Also, you will need to default this attribute to an empty hash in the attributes. https://github.com/cerner/cerner_splunk/blob/stable/attributes/_configure.rb.

rob-sch1: I don't think there's any harm in having an empty log-local.cfg (splunk just falls back to the regular log.cfg for anything not defined in log-local), but I can add the empty check if we want.

Ah, thanks for the tip about the attributes.

rob-sch1: Addressed in these two commits:
https://github.com/cerner/cerner_splunk/pull/191/commits/9067447497130d498df2944c87233eec4e7fd019
https://github.com/cerner/cerner_splunk/pull/191/commits/129f676ebf640bdb4e4a69dc3c8ca2caa34c6eb9

nmattam: > I don't think there's any harm in having an empty log-local.cfg

I don't think it hurts to have an empty log-local.cfg either, but I think the problem here is that the next time when someone runs chef-client with this change, it will end up creating the `log-local.cfg` file (empty or not), causing Splunk to restart (because we are touching the file marker). So even if someone has not configured these logging attributes, this change will end up forcing a restart, which is not the expected behavior.

gravesb:
Merged in https://github.com/cerner/cerner_splunk/commit/fe841a7d3cdbe4d980a72cfc1953732136a5f8b8.

PR #194, opened 2019-08-09
Update boot start scripts

gravesb:
Updated to follow the latest documentation:
* https://docs.splunk.com/Documentation/Splunk/latest/Admin/ConfigureSplunktostartatboottime#Enable_boot-start_as_a_non-root_user
* https://docs.splunk.com/Documentation/Splunk/7.3.1/Admin/RunSplunkassystemdservice#Configure_systemd_using_enable_boot-start

While I was testing, I ran into a number of other largely unrelated issues with the vagrant setup running under systemd.  Also, the windows install has been completely broken for some time, so I fixed that too.

### DIFF
--- a/recipes/_start.rb
+++ b/recipes/_start.rb
@@ -18,20 +18,37 @@ execute command do
   not_if { File.exist?(node['splunk']['systemd_file_location']) }
 end
 
-ruby_block 'insert ulimit' do
+ruby_block 'update-initd-file' do
   block do
     file = Chef::Util::FileEdit.new(init_file_path)
     file.insert_line_after_match(/^RETVAL=\d$/, ulimit_command)
+    file.insert_line_after_match(/^#{ulimit_command}$/, "USER=#{node['splunk']['user']}")
+    file.search_file_replace(%r{"[$\w\/]+\/bin\/splunk" start --no-prompt --answer-yes}, "su - ${USER} -c '\\0'")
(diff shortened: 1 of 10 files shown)


recipes/_start.rb
```diff
@@ -41,12 +58,12 @@
   notifies :start, 'service[splunk]', :immediately
 end
 
-# Added in because the first time splunk is started using chef and systemd the .pid file is owned by root which causes issues. It is automatically correct when splunk is restarted.
-execute 'correct permissions' do
-  command "chown -RP #{node['splunk']['user']}:#{node['splunk']['group']} #{node['splunk']['home']}"
-  action :run
-  not_if { platform_family?('windows') }
-  only_if { Gem::Version.new(node['splunk']['package']['version']) >= Gem::Version.new('7.2.2') && node['platform_version'].to_i == 7 }
+# The first time splunk is started on linux using chef the .pid file is owned by root which causes issues.
```
gravesb: @som3guy How come this was added?  I tried removing it and everything still seemed to work fine, but maybe I'm just not running the same test scenario you were.

som3guy: It was added during testing the deployment of Splunk on the project I have been working on for the last couple of months in dev.

I am not certain what caused it exactly, my guess is somehow splunk was starting as the root user at some point in the chef-client run. It would no longer restart because the restart was happening as the splunk user.

It was happening on a new installation of EL 7.4 and Splunk version 7.2.6

PR #203, opened 2019-10-15
Fix cerner_splunk_sh_cluster remove action

gravesb:
When executing the remove action on the cerner_splunk_sh_cluster resource, the following error occurs:
```
================================================================================
Error executing action `remove` on resource 'cerner_splunk_sh_cluster[remove SH from SHC]'
================================================================================

NoMethodError
-------------
undefined method `management_host' for Chef::Resource::Execute

Cookbook Trace:
---------------
/var/chef/cache/cookbooks/cerner_splunk/providers/sh_cluster.rb:43:in `block (2 levels) in class_from_file'
/var/chef/cache/cookbooks/cerner_splunk/providers/sh_cluster.rb:39:in `block in class_from_file'
```


spec/unit/resources/sh_cluster.rb
```diff
@@ -2,15 +2,15 @@
```
gravesb: For some reason that I haven't been able to figure out yet, this test doesn't get executed when you just run `rspec`.  The only way I've gotten it to run is if I do it explicitly via `rspec spec/unit/resources/sh_cluster.rb`.  So the test has been failing for a while, we just didn't notice it.

gravesb: I figured it out.  When you run `rspec` it only looks for files that end with `_spec.rb`.  The sh_cluster spec file was just named wrong, so it wasn't getting picked up.

PR #212, opened 2020-02-03
Add Support for Chef-client 15

tderanek:
# Details
* Adding Support for Chef-client 15
  * This cookbook's chef-client restriction is impacting other cookbooks that leverage it. The latest Chef-client version cannot be leveraged by any cookbooks that consume share a run-list with the cerner_splunk cookbook
* We would like to lift this restriction so that consuming cookbooks will be free to leverage the latest chef-client version

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -6,12 +6,12 @@ maintainer_email 'splunk@cerner.com'
 license          'Apache-2.0'
 description      'Installs/Configures Splunk Servers and Forwarders'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version          '2.35.0'
+version          '2.36.0'
 
 source_url       'https://github.com/cerner/cerner_splunk'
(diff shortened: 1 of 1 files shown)


gravesb:
If I modify the vagrant file to use Chef 15, I get the following error when spinning up one of the vagrant integration testing boxes:
```
==> s_standalone: Running chef-client...
==> s_standalone: Chef Infra Client cannot execute without accepting the license
Chef never successfully completed! Any errors should be visible in the
output above. Please fix your recipes so that they properly complete.
```

That will need to be resolved before we merge this.  In addition, I would expect a number of other changes in the Gemfile and .travis.yml to allow for the unit tests to be executed against ruby 2.6.5 and Chef 15.  A simple update to the metadata is not sufficient, the tests need to be updated accordingly to ensure the cookbook actually works with it.

gravesb:
See #213

PR #213, opened 2020-02-14
Add Support for Chef-client 15

snehitgajjar:
* Adding Support for Chef-client 15
* This cookbook's chef-client restriction is impacting other cookbooks that leverage it. The latest Chef-client version cannot be leveraged by any cookbooks that consume share a run-list with the cerner_splunk cookbook
* We would like to lift this restriction so that consuming cookbooks will be free to leverage the latest chef-client version

@gravesb please review and test it locally.

PR #216, opened 2020-03-04
Remove unnecessary windows_password configs

gravesb:
It turns out, we were requiring users to configure a value that was never actually used.  See the comments inline for additional information.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -351,13 +351,16 @@ Vagrant.configure('2') do |config|
     cfg.vm.guest = :windows
     # config below prevents the installation of latest Chef on the box.
     # Reference: https://github.com/chef/vagrant-omnibus/issues/118
-    config.omnibus.install_url = 'https://packages.chef.io/files/stable/chef/14.5.27/windows/2012r2/chef-client-14.5.27-1-x64.msi'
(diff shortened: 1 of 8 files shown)


Vagrantfile
```diff
@@ -351,13 +351,16 @@ Vagrant.configure('2') do |config|
...
     # Reference: https://github.com/chef/vagrant-omnibus/issues/118
-    config.omnibus.install_url = 'https://packages.chef.io/files/stable/chef/14.5.27/windows/2012r2/chef-client-14.5.27-1-x64.msi'
+    # TODO: figure out how to install chef 15 on a windows box
+    cfg.omnibus.install_url = 'https://packages.chef.io/files/stable/chef/14.5.27/windows/2012r2/chef-client-14.5.27-1-x64.msi'
     cfg.vm.provider :virtualbox do |vb|
       vb.customize ['modifyvm', :id, '--memory', 1024]
     end
     cfg.vm.provision :chef_client do |chef|
       chef_defaults chef, :f_win2012r2, 'splunk_standalone'
+      chef.arguments = ''
```
gravesb: This was missed in #213 and I haven't been able to get it working with Chef 15 yet. So for now I left it on 14 so I could test the other changes I actually wanted to make.

PR #222, opened 2020-08-14
Loosen version restriction on chef-vault

gravesb:
When running on Chef Client 15.13.8 and above, the following deprecation warning is present:
```
Resource chef_vault_secret needs `provides :chef_vault_secret` in addition to `resource_name :chef_vault_secret` declaration
Please see https://docs.chef.io/deprecations_resource_name_without_provides/ for further details and information on how to correct this problem.
```
This has been corrected in chef-vault 4.0.3, so we need to open up the version restriction to allow for newer versions of chef-vault to be used with cerner_splunk.

I didn't put an upper limit because I don't believe the chef-vault cookbook will be getting much more development, if any, because it is now included in the [Chef Infra client package as of Chef 16](https://docs.chef.io/release_notes/#chef-vault-functionality-out-of-the-box). Once we drop support for older chef clients (< 16), we can remove it from the metadata entirely.

### DIFF
--- a/libraries/splunk_app.rb
+++ b/libraries/splunk_app.rb
@@ -325,7 +325,7 @@ class Chef
         proc_sym = data['proc'].to_sym
         data = symbolize_keys(data).reject { |k, _| k == :proc }
         arguments = context.merge data
-        source_module.send proc_sym, arguments
+        source_module.send proc_sym, **arguments
       end
 
       def insert_procs(filename, contents)
--- a/metadata.rb
+++ b/metadata.rb
@@ -13,7 +13,7 @@ issues_url       'https://github.com/cerner/cerner_splunk/issues'
 
 chef_version     '>= 12.7', '< 17'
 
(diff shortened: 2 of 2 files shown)


libraries/splunk_app.rb
```diff
@@ -325,7 +325,7 @@ def hash_to_proc(source_module, data, context = {})
         proc_sym = data['proc'].to_sym
         data = symbolize_keys(data).reject { |k, _| k == :proc }
         arguments = context.merge data
-        source_module.send proc_sym, arguments
+        source_module.send proc_sym, **arguments
```
gravesb, line 328: While doing a bit more testing, I noticed this warning showing up in the chef output under Chef 16/Ruby2.7.1:
```
/var/chef/cache/cookbooks/cerner_splunk/libraries/splunk_app.rb:328: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```
This change resolves that warning.

PR #226, opened 2020-10-15
Fix install idempotency on windows

kimbernator:
The splunk_home attribute was being set based purely on the base_name attribute, which is not correct on windows. By calling the existing library function `CernerSplunk.installed_package_name`, the actual home directory can be found and the existence of the manifest verified. Prior to this, installation on windows would be attempted every time `_install` ran, resulting in success on the first run and failure on any subsequent runs.

Additionally, the same is now called on the package name for the `windows_package` resource so it can be properly checked against the uninstall registry keys, and the version no longer appends the build ID as that only a characteristic of linux package management.

### DIFF
--- a/recipes/_install.rb
+++ b/recipes/_install.rb
@@ -15,7 +15,7 @@ node.default['splunk']['package']['name'] = "#{nsp['base_name']}-#{nsp['version'
 node.default['splunk']['package']['file_name'] = "#{nsp['name']}#{nsp['file_suffix']}"
 node.default['splunk']['package']['url'] =
   "#{nsp['base_url']}/#{nsp['download_group']}/releases/#{nsp['version']}/#{nsp['platform']}/#{nsp['file_name']}"
-node.default['splunk']['home'] = CernerSplunk.splunk_home(node['platform_family'], node['kernel']['machine'], nsp['base_name'])
+node.default['splunk']['home'] = CernerSplunk.splunk_home(node['platform_family'], node['kernel']['machine'], CernerSplunk.installed_package_name(node['platform_family'], node['splunk']['package']['base_name']))
 node.default['splunk']['cmd'] = CernerSplunk.splunk_command(node)
 
 service = CernerSplunk.splunk_service_name(node['platform_family'], nsp['base_name'])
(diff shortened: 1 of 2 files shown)


recipes/_install.rb
```diff
@@ -15,7 +15,7 @@ def nsp
 node.default['splunk']['package']['file_name'] = "#{nsp['name']}#{nsp['file_suffix']}"
 node.default['splunk']['package']['url'] =
   "#{nsp['base_url']}/#{nsp['download_group']}/releases/#{nsp['version']}/#{nsp['platform']}/#{nsp['file_name']}"
-node.default['splunk']['home'] = CernerSplunk.splunk_home(node['platform_family'], node['kernel']['machine'], nsp['base_name'])
+node.default['splunk']['home'] = CernerSplunk.splunk_home(node['platform_family'], node['kernel']['machine'], CernerSplunk.installed_package_name(node['platform_family'], node['splunk']['package']['base_name']))
```
gravesb, line 18: When I test this on the f_win2012r2 vagrant box, the forwarder is still installed under `C:\Program Files\splunkforwarder`, so the existing values seem correct.  Does this needs to flex based on windows version?  I'm trying to get a newer version of windows for the vagrant testing, but having some issues finding a working vagrant box.

gravesb: Nevermind..forgot we were setting the INSTALLDIR flag during the windows_package install.  If I don't set the INSTALLDIR, the default appears to actually be SplunkUniversalForwarder as opposed to just UniversalForwarder.  So what does changing from `splunkforwarder` to `UniversalForwarder` for the install dir actually do?

kimbernator: It does seem like INSTALLDIR being set would prevent this from being an issue, but we're seeing all of our nodes install to C:\program files\UniversalForwarder instead of splunkforwarder. When chef runs for a second time, it bombs on the `windows_package` resource below because `manifest_missing` goes to check C:\program files\splunkforwarder for the manifest file.

I may need to investigate why it's installing to a that directory instead of the one specified in INSTALLDIR

kimbernator: Okay, it looks like this may not be the fix I'm after, nor would it provide any meaningful benefit right now. I'll close this out.

PR #233, opened 2021-02-22
Change how default sslPassword is determined

HCarr64:
Using certs with a different password then what is determined by cerner_splunk would be overridden by what is placed by default in `/$SPLUNK_HOME/etc/system/local/server.conf`. Assigning that to an attribute allows us to change the sslPassword.


attributes/_configure.rb
```diff
@@ -53,3 +53,5 @@
 
 # Default systemd file location based on default systemd unit file name
 default['splunk']['systemd_file_location'] = '/etc/systemd/system/splunk.service'
+
+default['splunk']['sslConfig']['sslPassword'] = 'password'
```
gravesb, line 57: I think this should go somewhere under the splunk/config hash for the sake of consistency.  maybe splunk/config/server/sslconfig/sslpassword or just splunk/config/sslpassword? Or somewhere in the middle...I'm open to suggestion.

gravesb (commented):
Please also update docs/attributes.md and increment the version in metadata.rb.

gravesb (approved):

srikanthreddy5 (approved):

gravesb:
Thanks for the contribution.

PR #244, opened 2021-05-03
Allow for configurable Splunk user home directory

gravesb:
For consistency, some systems may want all users to have the same base home directory.  The splunk user created by the splunk package installation has the home directory defaulted to SPLUNK_HOME (/opt/splunk).  This enhancement keeps that as the default behavior but allows consumers to configure a different home directory if they so choose.

### DIFF
--- a/attributes/_configure.rb
+++ b/attributes/_configure.rb
@@ -42,7 +42,7 @@ default['splunk']['boot_start_args'] =
     case node['platform_version'].to_i
     when 6
       '-systemd-managed 0'
-    when 7
+    when 7, 8
       '-systemd-managed 1 -systemd-unit-file-name splunk'
     else
       '-systemd-managed 0'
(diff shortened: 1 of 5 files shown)


attributes/_configure.rb
```diff
@@ -42,7 +42,7 @@
     case node['platform_version'].to_i
     when 6
       '-systemd-managed 0'
-    when 7
+    when 7, 8
```
gravesb, line 45: I'm still not exactly sure what changed to require this. A few weeks ago, the vagrant boxes came up fine on centos 8.x with initd, and some of my tests to switch to systemd were actually failing (though I don't recall the exact error).  Now, trying to use initd fails and it works fine with systemd...

som3guy (approved):

srikanthreddy5 (approved):

PR #246, opened 2021-05-24
Allow configuration of auth header for lookup urls

gravesb:
This enhancement allows an authentication header to be configured for lookup urls in case that lookup is hidden behind an auth wall.  We may be able to do something similar for apps, but I haven't had a need to do so yet.

I tried testing the splunk_app changes, but since it's written in such an old style of lwrp it's extremely difficult to unit test it.  I think it needs to be rewritten in the form of a more modern lwrp and at that point we can create some decent unit tests for it.


.rubocop.yml
```diff
@@ -22,10 +19,19 @@ Lint/RaiseException:
...
+
+Metrics/BlockLength:
+  Enabled: false
+
 Metrics/MethodLength:
   Max: 40
 
-Metrics/AbcSize:
+Metrics/PerceivedComplexity:
+  Max: 9
```
gravesb, line 32: This is the only real change here.  The other stuff is just alphabetizing.

som3guy (approved):

srikanthreddy5 (approved):

PR #252, opened 2022-01-20
Update default version to 8.1.8

gravesb:
When I did the last upgrade to 8.1.7.2, I tested the full package install but I forgot to test the UF.  As it turns out, since the UF wasn't affected by the log4j vulns, they didn't release an 8.1.7.2 version of it. For the UF, there was 8.1.7 and then 8.1.8.  Updating the default to 8.1.8 so consumers of the UF don't have to override the version attribute to get a successful chef run.

### DIFF
--- a/attributes/_install.rb
+++ b/attributes/_install.rb
@@ -13,8 +13,8 @@ default['splunk']['external_config_directory'] =
     '/etc/splunk'
   end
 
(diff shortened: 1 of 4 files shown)


som3guy (approved):

srikanthreddy5 (approved):

srikanthreddy5:
+1

PR #265, opened 2023-01-26
Add an ignore attribute to allow installing 2 splunk instances.

HCarr64:
Add an ignore attribute to allow installing 2 splunk instances.

Added ignore_already_installed_instance attribute default false.
Added ^ to the forwarder recipe.
Added ignore_already_installed_instance as true to s_standalone environment.

Tested this by running chef and s_standalone normally and then running cerner_splunk::forwarder on s_standalone.

https://gist.github.com/HCarr64/2e7fbc6f22b51157d457d133e77b043e

Vagrant Tests:
chef & s_standalone: https://gist.github.com/HCarr64/2e7fbc6f22b51157d457d133e77b043e
s1 & s2: https://gist.github.com/HCarr64/9066d1cb23795eebc8c10b9495392e8f
c1: https://gist.github.com/HCarr64/b65791855eb841f1628dc98d58f67e97
c2: https://gist.github.com/HCarr64/20a2efde542782cf460eef83c5e5dc87 I've never been able to get c2_newnode to run correctly.
f: https://gist.github.com/HCarr64/ec895cc9ee2501dad13ff0aa7c742dc8
s_license: https://gist.github.com/HCarr64/5f98316122449745d1b6f0029f69e757

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -24,6 +24,7 @@ Vagrant.require_version '>= 2.2.5'
   s2_search:    { ip: '192.168.56.36', hostname: 's2.search.splunk', ports: { 8017 => 8000, 8107 => 8089 } },
   s2_slave1:    { ip: '192.168.56.37', hostname: 's2.slave01.splunk', ports: { 8018 => 8000, 8108 => 8089 } },
   s2_slave2:    { ip: '192.168.56.38', hostname: 's2.slave02.splunk', ports: { 8019 => 8000, 8109 => 8089 } },
+  sf_standalone:{ ip: '192.168.56.39', hostname: 'splunk3', ports: { 8020 => 8000, 8110 => 8089 } },
   f_default:    { ip: '192.168.56.50', hostname: 'default.forward', ports: { 9090 => 8089 } },
   f_debian:     { ip: '192.168.56.51', hostname: 'debian.forward', ports: { 9091 => 8089 } },
   f_heavy:      { ip: '192.168.56.52', hostname: 'heavy.forward', ports: { 9092 => 8089 } },
@@ -265,9 +266,11 @@ Vagrant.configure('2') do |config|
 
   config.vm.define :s_standalone do |cfg|
     cfg.vm.provision :chef_client do |chef|
-      chef_defaults chef, :s_standalone, 'splunk_standalone'
+#      chef_defaults chef, :s_standalone, 'splunk_standalone'
+      chef_defaults chef, :s_standalone, 'splunkf_standalone'
       chef.add_recipe 'cerner_splunk_test::install_libarchive'
-      chef.add_recipe 'cerner_splunk::server'
+#    chef.add_recipe 'cerner_splunk::server'
+     chef.add_recipe 'cerner_splunk::forwarder'
     end
     network cfg, :s_standalone
   end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -24,7 +24,6 @@ Vagrant.require_version '>= 2.2.5'
   s2_search:    { ip: '192.168.56.36', hostname: 's2.search.splunk', ports: { 8017 => 8000, 8107 => 8089 } },
   s2_slave1:    { ip: '192.168.56.37', hostname: 's2.slave01.splunk', ports: { 8018 => 8000, 8108 => 8089 } },
   s2_slave2:    { ip: '192.168.56.38', hostname: 's2.slave02.splunk', ports: { 8019 => 8000, 8109 => 8089 } },
-  sf_standalone:{ ip: '192.168.56.39', hostname: 'splunk3', ports: { 8020 => 8000, 8110 => 8089 } },
   f_default:    { ip: '192.168.56.50', hostname: 'default.forward', ports: { 9090 => 8089 } },
   f_debian:     { ip: '192.168.56.51', hostname: 'debian.forward', ports: { 9091 => 8089 } },
   f_heavy:      { ip: '192.168.56.52', hostname: 'heavy.forward', ports: { 9092 => 8089 } },
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -266,11 +266,13 @@ Vagrant.configure('2') do |config|
 
   config.vm.define :s_standalone do |cfg|
     cfg.vm.provision :chef_client do |chef|
-#      chef_defaults chef, :s_standalone, 'splunk_standalone'
-      chef_defaults chef, :s_standalone, 'splunkf_standalone'
+      chef_defaults chef, :s_standalone, 'splunk_standalone'
+# This is used to test installing the forwarder ontop of a node that already has splunk installed. Only have 1 chef_defaults line active.
+#      chef_defaults chef, :s_standalone, 'splunkf_standalone'
       chef.add_recipe 'cerner_splunk_test::install_libarchive'
-#    chef.add_recipe 'cerner_splunk::server'
-     chef.add_recipe 'cerner_splunk::forwarder'
+      chef.add_recipe 'cerner_splunk::server'
+# This is used to test installing the forwarder ontop of a node that already has splunk installed. Only have splunk_server OR forwarder active.
+#      chef.add_recipe 'cerner_splunk::forwarder'
     end
     network cfg, :s_standalone
   end
--- a/libraries/recipe_utils.rb
+++ b/libraries/recipe_utils.rb
@@ -144,9 +144,7 @@ module CernerSplunk # rubocop:disable Metrics/ModuleLength
       return 'SplunkForwarder' if package_base_name == 'splunkforwarder'
       return 'Splunkd' if package_base_name == 'splunk'
     end
-      return systemd_unit_file_name
-    #return 'splunkforwarder' if package_base_name == 'splunkforwarder'
-    #'splunk'
+    return systemd_unit_file_name
   end
 
   # Validates that the splunk.secret file either does not exist or has the same value that's currently configured.
--- a/libraries/recipe_utils.rb
+++ b/libraries/recipe_utils.rb
@@ -103,7 +103,7 @@ module CernerSplunk # rubocop:disable Metrics/ModuleLength
   # Returns the installed package name based on platform and package base name.
   # Written because Windows is dumb
   def self.installed_package_name(platform_family, package_base_name, systemd_unit_file_name)
-    return package_base_name unless systemd_unit_file_name != '' unless platform_family == 'windows'
+    return package_base_name unless platform_family == 'windows'
 
     # Windows package names
     return 'Splunk Enterprise' if package_base_name == 'splunk'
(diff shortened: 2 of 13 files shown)


recipes/forwarder.rb
```diff
@@ -11,7 +11,7 @@
 node.default['splunk']['package']['base_name'] = 'splunkforwarder'
 node.default['splunk']['package']['download_group'] = 'universalforwarder'
 
-fail 'Different Splunk artifact already installed on node. Failing as an unsupported install' if CernerSplunk.separate_splunk_installed?(node)
+fail 'Different Splunk artifact already installed on node. Failing as an unsupported install' if CernerSplunk.separate_splunk_installed?(node) && node['splunk']['ignore_already_installed_instance'] == false
```
gravesb, line 14:
```suggestion
fail 'Different Splunk artifact already installed on node. Failing as an unsupported install' if CernerSplunk.separate_splunk_installed?(node) && !node['splunk']['ignore_already_installed_instance']
```

gravesb (commented):
I'm a little surprised the second chef run succeeded even though the splunkforwarder failed to start.  Can you setup the test so it actually works and both splunk services start up?  The test configs can be helpful as examples for others trying to configure the cookbook, so I would like to have a fully working setup.  Might need to add some additional chef config, or a new environment so chef can successfully install and start both Splunk installs.  If it's easier, you might consider creating a whole new box for it too.

And just to clarify, it will have to be two separate chef runs to accomplish this right?  I don't think there's any way to install both packages in a single run, at least not without some major refactoring.

HCarr64:
> And just to clarify, it will have to be two separate chef runs to accomplish this right? I don't think there's any way to install both packages in a single run, at least not without some major refactoring.

Yes, that is correct

HCarr64:
This currently works in linux, but more testing needs to be done for existing setups.

Vagrantfile
```diff
@@ -24,6 +24,7 @@ Vagrant.require_version '>= 2.2.5'
   s2_search:    { ip: '192.168.56.36', hostname: 's2.search.splunk', ports: { 8017 => 8000, 8107 => 8089 } },
   s2_slave1:    { ip: '192.168.56.37', hostname: 's2.slave01.splunk', ports: { 8018 => 8000, 8108 => 8089 } },
   s2_slave2:    { ip: '192.168.56.38', hostname: 's2.slave02.splunk', ports: { 8019 => 8000, 8109 => 8089 } },
+  sf_standalone:{ ip: '192.168.56.39', hostname: 'splunk3', ports: { 8020 => 8000, 8110 => 8089 } },
```
gravesb, line 27: it doesn't look like this is used, so it can be deleted.

libraries/recipe_utils.rb
```diff
@@ -102,8 +102,8 @@ def self.opposite_package_name(package_base_name)
 
   # Returns the installed package name based on platform and package base name.
   # Written because Windows is dumb
-  def self.installed_package_name(platform_family, package_base_name)
-    return package_base_name unless platform_family == 'windows'
+  def self.installed_package_name(platform_family, package_base_name, systemd_unit_file_name)
+    return package_base_name unless systemd_unit_file_name != '' unless platform_family == 'windows'
```
gravesb, line 106: what is this change for?  I don't think this is needed, but if it is then you shouldn't have two unless statements.  It would be `unless systemd_unit_file_name != '' || platform_family == 'windows'`.

libraries/recipe_utils.rb
```diff
@@ -139,13 +139,14 @@ def self.use_splunk_start_command?(node, previous_splunk_version)
...
+  def self.splunk_service_name(platform_family, package_base_name, systemd_unit_file_name)
     if platform_family == 'windows'
       return 'SplunkForwarder' if package_base_name == 'splunkforwarder'
       return 'Splunkd' if package_base_name == 'splunk'
     end
-
-    'splunk'
+      return systemd_unit_file_name
+    #return 'splunkforwarder' if package_base_name == 'splunkforwarder'
+    #'splunk'
```
gravesb, line 149: fix indentation and remove commented code.

recipes/_generate_password.rb
```diff
@@ -8,6 +8,8 @@
 
 return if node['splunk']['free_license'] && node['splunk']['node_type'] != :forwarder
 
+return if node['splunk']['ignore_already_installed_instance'] == true
```
gravesb, line 11: why are we skipping this step?  Under normal circumstances we do this for both UFs and full package installs.

recipes/_start.rb
```diff
@@ -11,8 +11,10 @@
 
 # We want to always ensure that the boot-start script is in place on non-windows platforms
 package_version = Gem::Version.new(node['splunk']['package']['version'])
-command = "#{node['splunk']['cmd']} enable boot-start -user #{node['splunk']['user']}"
+command = "#{node['splunk']['cmd']} enable boot-start -user #{node['splunk']['user']} -systemd-unit-file-name #{node['splunk']['systemd_unit_file_name']}"
```
gravesb, line 14: It looks like this argument was introduced in 7.2.2.  I think the systemd-unit-file-name argument needs to be appended down on line 18 instead.

recipes/_start.rb
```diff
@@ -11,8 +11,10 @@
...
 command += " -group #{node['splunk']['group']}" if package_version >= Gem::Version.new('7.3.0')
+#####
+#command +=
```
gravesb, line 17: remove commented code.

gravesb (changes requested):

recipes/_generate_password.rb
HCarr64: It caused issue with the other instance of splunk that is running. It would change the password on file but not actually change the password running in splunk. 

Now that I think about it we would have the same issue when a normal chef would be ran and then the universalforwarder would have a wrong password. I'll have to change this up.

Vagrantfile
HCarr64: Removed: https://github.com/cerner/cerner_splunk/pull/265/commits/3c791c8b0b39737525dc63252bed91702aa74c01

libraries/recipe_utils.rb
HCarr64: Removed: https://github.com/cerner/cerner_splunk/pull/265/commits/8f3f110b8b4b23558dafe16fa9d191478f576e58

recipes/_start.rb
HCarr64: Removed: https://github.com/cerner/cerner_splunk/pull/265/commits/6d21496cf85b6f8ddfdc94851e4ce32adce0c505

recipes/_generate_password.rb
HCarr64: https://github.com/cerner/cerner_splunk/pull/265/commits/2446ced75d2da233b1901d9109f92917bce8cf1f

This should fix this issue. Gist of tests https://gist.github.com/HCarr64/2e7fbc6f22b51157d457d133e77b043e

Vagrantfile
```diff
@@ -266,8 +266,12 @@ Vagrant.configure('2') do |config|
   config.vm.define :s_standalone do |cfg|
     cfg.vm.provision :chef_client do |chef|
       chef_defaults chef, :s_standalone, 'splunk_standalone'
+# This is used to test installing the forwarder ontop of a node that already has splunk installed. Only have 1 chef_defaults line active.
+#      chef_defaults chef, :s_standalone, 'splunkf_standalone'
       chef.add_recipe 'cerner_splunk_test::install_libarchive'
       chef.add_recipe 'cerner_splunk::server'
+# This is used to test installing the forwarder ontop of a node that already has splunk installed. Only have splunk_server OR forwarder active.
```
gravesb, line 273: For consistency, we should probably add a blurb to the docs/vagrant.md file around how to test installing a forwarder along side the full server.

recipes/_generate_password.rb
```diff
@@ -10,7 +10,12 @@
 
 require 'securerandom'
 
-password_file = File.join node['splunk']['external_config_directory'], 'password'
+# For a node that has multiple splunk instances, create a separate password file.
+password_file = if node['splunk']['ignore_already_installed_instance'] == true
```
gravesb, line 14: minor, shouldn't need the `== true` here.  I'm surprised rubocoop didn't flag that.

recipes/_start.rb
```diff
@@ -13,7 +13,8 @@
 package_version = Gem::Version.new(node['splunk']['package']['version'])
 command = "#{node['splunk']['cmd']} enable boot-start -user #{node['splunk']['user']}"
 command += " -group #{node['splunk']['group']}" if package_version >= Gem::Version.new('7.3.0')
-command += " #{node['splunk']['boot_start_args']}" if package_version >= Gem::Version.new('7.2.2')
+command += " #{node['splunk']['boot_start_args']} -systemd-unit-file-name #{node['splunk']['systemd_unit_file_name']}" if node['platform_version'].to_i != 6 && (package_version >= Gem::Version.new('7.2.2'))
+command += " #{node['splunk']['boot_start_args']}" if node['platform_version'].to_i == 6 && (package_version >= Gem::Version.new('7.2.2'))
```
gravesb, line 17:
```suggestion
command += " #{node['splunk']['boot_start_args']}" if package_version >= Gem::Version.new('7.2.2')
command += "  -systemd-unit-file-name #{node['splunk']['systemd_unit_file_name']}" if node['platform_version'].to_i > 6 && (package_version >= Gem::Version.new('7.2.2'))
```

Do we also need to worry about the platform family here (i.e. windows)?  Previously, we only set systemd-managed 1 and the systemd-unit-file in the default boot start args when the platform family was rhel,fedora,or amazon and the platform_version was 7 or 8.  This only checks the version.

Vagrantfile
HCarr64: Added: https://github.com/cerner/cerner_splunk/pull/265/commits/75111f9aeb55cce020b4a7a2d71e93adc82fc1db

recipes/_generate_password.rb
HCarr64: Removed: https://github.com/cerner/cerner_splunk/pull/265/commits/b82406d1b35f6d5af0f356b65b48a4bad25c628b

Reran rubocop as a sanity check, still didn't catch it.

recipes/_start.rb
HCarr64: When the platform version is 6 it doesn't want to use systemd according to rspec.

gravesb: nevermind about the windows bit, the guard on line 20 takes care of that.  But I still think the code suggestion I provided is a bit cleaner since it doesn't repeat the boot_start_args.

HCarr64: Cleaned: https://github.com/cerner/cerner_splunk/pull/265/commits/96382c0de99b55664daaf8e4da926ea4029fda4e

2 splunk installs tests and rspec still passes as well.

